PR #2, opened 2016-04-08
Slider transitions

rusev:


gyoshev:
Squashed & merged as 83efc9eb9848b66

PR #6, opened 2016-04-13
refactor(Animation): remove unnecessary comments

ggkrustev:
The comments that were used to force the root rule generation turned out to be unnecessary. It seems
that CDN generation skips comments before css generation.


gyoshev:
`DO NOT MOVE` was picking a fight

PR #9, opened 2016-05-04
feat(animation): add an animation child container styling

ggkrustev:
The class will be rendered to the `k-animation-container` children


ggkrustev:
@gyoshev  Do you think that the naming of the CSS class is ok?

PR #12, opened 2016-05-20
fix(slider): add d.ts file for slider

KirilNN:
Adding the missing d.ts file, so the module can be used in Angular components as well.


gyoshev:
At this time we have one .d.ts per package, so the file should be defined as `styles/packages/inputs.d.ts`.

PR #20, opened 2016-08-26
Scrollview styles for Angular theme

theOrlin:
First time creating styles for Kendo Angular 2 widget. Want to make sure everything is OK.


styles/scrollview/_layout.scss
```diff
@@ -0,0 +1,132 @@
...
+        height: 100%;
+        cursor: default;
+
+        img,
+        li {
+            -webkit-user-select: none;
+            -khtml-user-select: none;
+            -moz-user-select: none;
+            -o-user-select: none;
+            user-select: none;
```
gyoshev, line 24: The distribution files are post-processed with autoprefixer, so you can remove the prefixed properties.

PR #35, opened 2016-10-11
fix: chart default font-size; major/minor gridlines colors

inikolova:
Change chart default font-size to 14px. Change major / minor gridlens colors to match the design

### DIFF
--- a/styles/charts/_layout.scss
+++ b/styles/charts/_layout.scss
@@ -3,7 +3,7 @@
 $chart-tap-highlight-color: rgba(0, 0, 0, 0);
 $chart-selection-background: rgba(255, 255, 255, .01);
 
-$chart-font-size: 14px !default;
+$chart-font-size: $base-font-size !default;
 $chart-tooltip-font-size: .929em !default;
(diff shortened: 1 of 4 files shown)


styles/charts/_layout.scss
```diff
@@ -3,12 +3,22 @@
 $chart-tap-highlight-color: rgba(0, 0, 0, 0);
 $chart-selection-background: rgba(255, 255, 255, .01);
 
-$chart-font-size: 1em !default;
+$chart-font-size: 14px !default;
```
gyoshev: This appears to duplicate the font-size that is [set on components](https://github.com/telerik/kendo-theme-default/commit/7a7e4d20749b27e22943adc5b867abdd20e1bbef#diff-55c4e8f17f1e62da1f19895fb8fdb5d8R17). How about extracting both to a root-level variable in `_variables.scss`, so that it can be changed from a single place?

inikolova: Sounds good! I'll make this change

PR #46, opened 2016-11-10
Chart selection styling

inikolova:
Add styling for chart selection

### DIFF
--- a/styles/charts/_layout.scss
+++ b/styles/charts/_layout.scss
@@ -1,7 +1,6 @@
 @import '../base/layout';
 
 $chart-tap-highlight-color: rgba(0, 0, 0, 0);
-$chart-selection-background: rgba(255, 255, 255, .01);
 
 $chart-font-size: $base-font-size !default;
 $chart-tooltip-font-size: .929em !default;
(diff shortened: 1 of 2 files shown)


styles/charts/_theme.scss
```diff
@@ -110,6 +118,15 @@ $error-bars-background: rgba(0, 0, 0, .5) !default;
         background: $disabled-border-color;
     }
 
+    .k-marquee-color {
+        background-color: $chart-selection-marque-background;
+    }
+
+    .k-marquee-color {
+        filter: alpha(opacity=60);
```
gyoshev: We don't really need the `filter`, it's for IE5-7.

styles/charts/_layout.scss
```diff
@@ -117,11 +120,11 @@ $chart-title-font-size: 1.143em !default;
     }
 
     .k-leftHandle {
-        left: -4px;
+        left: -12px;
     }
 
     .k-rightHandle {
-        right: -4px;
+        right: -12px;
```
gyoshev: These seem to be related to the handle width. Use a variable to show the dependency.

PR #64, opened 2017-01-04
Fix upload styling

inikolova:
Fix broken appearance of the upload component

### DIFF
--- a/scss/upload/_layout.scss
+++ b/scss/upload/_layout.scss
@@ -1,4 +1,6 @@
 @include exports('upload/layout') {
+    $line-height-sm: (20 / 12) !default;
+
     html .k-upload {
         position: relative;
     }
@@ -114,6 +116,12 @@
         border-color: inherit;
         margin: 0;
 
+        .k-file-multiple,
+        .k-file-single {
+            display: block;
+            width: 100%;
+        }
+
         .k-file {
             padding: $padding-x;
             border-width: 0 0 1px;
@@ -147,7 +155,8 @@
             top: $padding-x + $padding-y;
 
             .k-button {
-                padding: 0;
+                padding: $padding-y-sm;
+                border-radius: 50%;
             }
         }
 
@@ -160,13 +169,11 @@
     .k-file-size,
     .k-file-validation-message,
     .k-file-information {
-        display: inline-block;
-        width: 100%;
+        display: block;
     }
 
     .k-file-name {
         position: relative;
-        vertical-align: middle;
         overflow: hidden;
         text-overflow: ellipsis;
         white-space: nowrap;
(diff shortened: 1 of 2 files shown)


scss/upload/_layout.scss
```diff
@@ -139,60 +147,58 @@
...
+
         .k-upload-status {
-            display: flex;
-            align-items: center;
-            position: relative;
-            margin-left: auto;
+            position: absolute;
+            line-height: 1em;
+            right: $padding-x;
+            top: $padding-x + $padding-y;
```
gyoshev: Why does horizontal padding influence the `top` position? Shouldn't that be `(multiplier) * $padding-y`, by any chance?

scss/upload/_layout.scss
```diff
@@ -210,8 +216,9 @@
     }
 
     .k-file-name-size-wrapper {
-        vertical-align: middle;
-        margin-left: 1em;
+        display: block;
+        margin-left: calc(#{$padding-x} + #{40px});
+        margin-right: calc(#{$padding-x} + #{16px + #{$padding-x-sm}});
```
gyoshev: What are `40px` and `16px`? How are these related to the surrounding code?

scss/upload/_theme.scss
```diff
@@ -11,18 +11,35 @@
     $upload-success-bg: $success !default;
     $upload-disabled-text: darken($widget-bg, 27%) !default;
     $upload-disabled-border: darken($widget-bg, 27%) !default;
+    $upload-focused-status-box-shadow: 0 0 0 2px darken($input-bg, 6%) !default;
```
gyoshev, line 14: How about naming this `$upload-focused-box-shadow`?

scss/upload/_layout.scss
```diff
@@ -1,43 +1,43 @@
...
     }
 
     .k-upload {
-        border-width: 1px;
-        border-style: solid;
-
-        .k-action-buttons {
-            margin: -6px 0 0;
+        .k-upload-button {
+            min-width: 6.643em;
```
gyoshev: Oddly specific. Why exactly this value?

scss/upload/_layout.scss
inikolova: There are as follows:
-40px - width of the uploaded file image;
-16px - clear icon width

scss/upload/_theme.scss
inikolova: This shadow is applied to the focused status icon only. Changing the var name to $upload-focused-box-shadow assumes that it will be applied to the Upload widhet itself. @gyoshev what do you think?

scss/upload/_layout.scss
inikolova: I agree. Done

scss/upload/_layout.scss
gyoshev: The icon size is available through the `$icon-size` variable, so using that will explain the intent.
The uploaded file image size seems computed from the `$padding-y-lg` and the width of `.k-file-extension-wrapper`. Using variables will continue to work after their values are changed, for example in a custom theme.

scss/upload/_layout.scss
inikolova: After further discussion with @yordanov I changed the upload min-width

scss/upload/_layout.scss
inikolova: You're right - I will make the change

Issue #65, opened 2017-01-05
Upload status icon position


Raisolution:
Status icon seems to be too close to the status text:
![success](https://cloud.githubusercontent.com/assets/730205/21673913/04acc3d2-d334-11e6-93c3-f61fd912a6e9.png)

![warning](https://cloud.githubusercontent.com/assets/730205/21673919/0ade5b8a-d334-11e6-8466-48ba7ce84b36.png)

Raisolution:
One more thing - the icon is not displayed during upload
![loading](https://cloud.githubusercontent.com/assets/730205/21674229/753115c6-d335-11e6-95f4-3c5efa582f6c.png)
The currently used CSS class is **k-i-loading**

PR #77, opened 2017-01-09
feat: add loading indicator. closes telerik/kendo-theme-default#65

inikolova:


scss/common/_base.scss
```diff
@@ -387,6 +386,53 @@
         opacity: .3;
     }
 
+    $loading-size: 12px !default;
+    $loading-speed: 2s !default;
+    $loading-shadow-width: 2px !default;
```
gyoshev: Is that the width of the animated stroke? If so, `$loading-stroke-width` may be more descriptive -- saying _what_ it is rather than _how_ it is implemented.

scss/common/_base.scss
```diff
@@ -387,6 +386,53 @@
...
+
+    .k-i-loading::after {
+        border-radius: 0 ( $loading-size + ($loading-shadow-width * 2)) ( $loading-size + ($loading-shadow-width * 2)) 0;
+        top: -($loading-shadow-width);
+        left: ($loading-size / 2);
+        transform-origin: 0 (($loading-size / 2) + $loading-shadow-width);
+        animation: loading-animation $loading-speed infinite ease;
+    }
+
+    @keyframes loading-animation {
```
gyoshev: I believe we should prefix this with `k-`, to avoid name clashes with animations on customer pages.

scss/common/_base.scss
```diff
@@ -387,6 +386,53 @@
         opacity: .3;
     }
 
+    $loading-size: 12px !default;
```
gyoshev: Is this bound to the `$icon-size`? Something like `$icon-size - 2 * $loading-shadow-width`?

scss/common/_base.scss
```diff
@@ -387,6 +386,53 @@
...
+    .k-i-loading::before {
+        border-radius: $loading-size + ($loading-shadow-width * 2) 0 0 $loading-size + ($loading-shadow-width * 2);
+        top: -($loading-shadow-width);
+        left: -($loading-shadow-width);
+        transform-origin: (($loading-size / 2) + $loading-shadow-width) (($loading-size / 2) + $loading-shadow-width);
+        animation: loading-animation $loading-speed infinite ease 1.5s;
+    }
+
+    .k-i-loading::after {
+        border-radius: 0 ( $loading-size + ($loading-shadow-width * 2)) ( $loading-size + ($loading-shadow-width * 2)) 0;
```
gyoshev: Helper variables will simplify these lines:

    $loading-outer-size: ($loading-size + ($loading-shadow-width * 2));
    $loading-half-size: $loading-outer-size / 2;
    // ^^^ this equals (($loading-size / 2) + $loading-shadow-width)

gyoshev (changes requested):

scss/common/_base.scss
inikolova: You are right - I will make the change

scss/common/_base.scss
inikolova: Yes, the $loading-size is related to the icon size. The suggested calculation sounds OK - I added it :)

scss/common/_base.scss
inikolova: You're right. Done!

scss/common/_base.scss
inikolova: Done

gyoshev (approved):

PR #81, opened 2017-01-10
chore(maskedtextbox): styles for maskedtextbox

joneff:
Styles for masked textbox.

Markup:

```
<kendo-maskedtextbox class="k-widget k-maskedtextbox">
  <input autocapitalize="off" autocomplete="off" autocorrect="off" class="k-input" />
</kendo-maskedtextbox> 
```

Question: since in jq-kendo masked textbox is simply `<input class="k-textbox" />`, should we also add the `k-textbox` class here and use those styles, while leaving the `k-maskedtextbox` class for scoping? Currently there is a slight duplication of styles e.g. longer selectors.

### DIFF
--- a/scss/input/_layout.scss
+++ b/scss/input/_layout.scss
@@ -5,38 +5,47 @@
     $textbox-shadow: 0 2px 2px 1px rgba(0, 0, 0, .06) !default;
     $fieldset-margin: 30px;
 
-    .k-textbox > input,
-    .k-input,
-    .k-textbox {
+    .k-textbox,
+    .k-maskedtextbox {
+        @include border-radius( $border-radius );
+        padding: $input-padding-y $input-padding-x;
+        width: 12.4em;
+        box-sizing: border-box;
         border-width: 1px;
         border-style: solid;
-        font-size: inherit;
+        outline: 0;
+        font: inherit;
+        line-height: $form-line-height;
+        display: inline-flex;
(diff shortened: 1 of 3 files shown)


scss/input/_layout.scss
```diff
@@ -5,38 +5,47 @@
...
     $fieldset-margin: 30px;
 
-    .k-textbox > input,
-    .k-input,
-    .k-textbox {
+    .k-textbox,
+    .k-maskedtextbox {
+        @include border-radius( $border-radius );
+        padding: $input-padding-y $input-padding-x;
+        width: 12.4em;
```
gyoshev, line 12: Is this default width shared with other components? If so, do we need a variable? If not, perhaps we don't need it at all (making inputs expand to 100%)?

gyoshev (commented):
Apart from the width comment, LGTM

scss/input/_layout.scss
joneff: The width is somewhat hardcoded in https://github.com/telerik/kendo-theme-default/blob/master/scss/common/_forms.scss. I've used the same width.

I was also wondering if I should put `k-maskedtextbox` _forms.

gyoshev: Seems like these should be together, or at least indicate that they share the logic. Perhaps we could use a placeholder / mixin for all these, so that the selectors are built based on the usage?

Issue #76, opened 2017-01-09
MultiSelct is missing classes for clear all button


KirilNN:
The element is rendered by default but is not visible on hover.

inikolova:
Done: efda3725c1c09ef671c1d

PR #82, opened 2017-01-10
feat: add styling for inputs clear all button

inikolova:
Add styling for inputs clear all button. Related to https://github.com/telerik/kendo-theme-default/issues/76

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -97,22 +97,29 @@
 
             .k-icon {}
         }
+    }
 
-
-        // Clear icon
-        .k-i-close {
+    .k-autocomplete,
+    .k-combobox > .k-dropdown-wrap,
+    .k-multiselect > .k-multiselect-wrap {
+        > .k-i-close {
             display: none;
-            align-self: center;
-        }
-
-        &.k-state-hover,
-        &.k-state-focused {
-            .k-i-close {
-                display: flex;
-                margin-right: $padding-y;
-                outline: 0;
-            }
+            position: absolute;
+            right: $padding-x;
+            top: $padding-y-lg;
         }
     }
 
+    .k-autocomplete.k-state-focused,
+    .k-autocomplete.k-state-hover,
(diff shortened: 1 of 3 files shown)


scss/multiselect/_layout.scss
```diff
@@ -19,13 +19,13 @@
...
 
-        > .k-i-close {
+        /*> .k-i-close {
             display: none;
             align-self: center;
             margin-left: auto;
             margin-right: $padding-y;
             outline: 0;
-        }
+        }*/
```
gyoshev: If it's not needed, delete it. It's available in the git history.

scss/common/_forms.scss
```diff
@@ -97,22 +97,29 @@
...
-                margin-right: $padding-y;
-                outline: 0;
-            }
+    .k-autocomplete.k-state-focused,
+    .k-autocomplete.k-state-hover,
+    .k-combobox>.k-dropdown-wrap.k-state-focused,
+    .k-combobox>.k-dropdown-wrap.k-state-hover,
+    .k-multiselect.k-state-focused>.k-multiselect-wrap,
+    .k-multiselect.k-state-hover>.k-multiselect-wrap,
+    .k-multiselect:hover>.k-multiselect-wrap {
```
gyoshev: Seems like this selector exploded. Looks like states are not applied consistently across these components. This can't be fixed right now, but maybe we should rethink how the states are applied.

gyoshev (changes requested):

scss/multiselect/_layout.scss
inikolova: Ooops - it appears I missed this commended rule. Deleted!

scss/common/_forms.scss
inikolova: I agree. Selectors - modified!

gyoshev: Very nice :+1:

gyoshev (approved):

Issue #52, opened 2016-12-08
v2 checklist


gyoshev:
After a quick pass through the online demos, it appears that the v2 branch is not ready for merging just yet. Here is a list of the things that we need to address to make an initial 2.0 release:

- [x] IN - grouping headers (combobox, ddl)
- [x] IN - datetimepicker icons alignment
-  datetimepicker icons should be stretched (currently clicks outside them can land on k-select and not open the picker) -> this is the default behavior of DateTimePicker. We should think about changing this behavior (breaking change)
- [x] IN - calendar selected today date background and foreground color are the same
- [x] IN - drop-down list icons are not vertically aligned
- [x] GYO - editor table width is missing
- [x] OR - editor icons color & layout
- [x] GYO - window k-overlay is not styled
- [x] IN - window icon colors
- [x] maskedtextbox layout
- [x] GYO - numerictextbox buttons
- [x] IN - slider - tooltip positioning is not right (probably due to animations?)
- [x] OR - upload - remove button is not right-aligned
- [x] OR - no styles for k-textbox (https://github.com/telerik/kendo-theme-default/issues/49)
- [x] GYO - switch, buttongroup have no styles due to km-styles (https://github.com/telerik/kendo/pull/6129)
- [x] IN - switch - labels do not get properly hidden
- [x] scrollview - appears to be missing layout
- [x] button demos - remove text from icon, review icons
- menu - no borders at all -> By design there aren't any borders
- [ ] menu - forecolor of expanded items
- [x] GYO - panelbar - k-image alignment
- [x] IN - tabstrip border exists between selection and content
- [ ] gantt, scheduler
- [x] IN - notification text in icons
- [x] OR - splitter drag styles
- [x] ZH - grid grouping icon text
- [ ] spreadsheet
- [x] IN - pager: add border by default, remove when nested
- [ ] mediaplayer - red borders
- [x] GYO - touch scroller is not styled (affects map)
- [x] GYO - dataviz - themes should be either defined in JSON for the kendo components, or the style functionality ported

- [ ] Kendo Notification close button is displaced with the Default skin [#78]( https://github.com/telerik/kendo-theme-default/issues/78)

Any contribution to the list (checking off, or adding new items) is welcome!

gyoshev:
v2 is out, remaining issues for the R1 release are tracked in another issue.

PR #84, opened 2017-01-10
fix: Editor toolbar styling #52

theOrlin:
This adds missing styles for the editor toolbar.

### DIFF
--- a/scss/editor/_layout.scss
+++ b/scss/editor/_layout.scss
@@ -1,5 +1,7 @@
 @include exports("editor/layout") {
 
+    $ct-cell-side-length: 20px !default;
+
     // Base
     .k-editor {
         border-collapse: separate;
@@ -80,6 +82,7 @@
                 border-radius: 0;
             }
         }
+
         .k-tool-group .k-state-disabled {
             display: none;
         }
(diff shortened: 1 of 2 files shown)


scss/editor/_layout.scss
```diff
@@ -153,10 +177,67 @@
...
+
+    .k-overlay {
+        position: fixed;
+        top: 0;
+        left: 0;
+        z-index: 10001;
+        width: 100%;
+        height: 100%;
+        background-color: #000;
+        filter: alpha(opacity=50);
```
gyoshev: No need for `filter`, given our browser support.

scss/editor/_theme.scss
```diff
@@ -1,12 +1,54 @@
...
+    }
+
+
+    .k-header.k-window-content.k-content {
+        @include appearance( header );
     }
 
+    .k-editor .k-editor-toolbar-wrap,
+    .k-ct-popup .k-editor-toolbar,
+    .k-header.k-window-content.k-content {
```
gyoshev: `.k-header.k-window-content.k-content` does not say much. Perhaps we need a more descriptive class, or to drop some of the selectors?

scss/editor/_layout.scss
```diff
@@ -89,29 +103,38 @@
...
+
+            .k-tool-text {
+                display: inline;
+            }
+        }
+
+        .k-ct-cell {
+            margin: $padding-y-sm / 2;
+            width: 20px;
+            height: 20px;
```
gyoshev: Is there a relationship between the `.k-ct-popup` width and padding, and the `.k-ct-cell` margin / width and height? It may be appropriate to calculate it. E.g. `.k-ct-popup { width: $cell-width * 10; }` means there will be 11 cells in the popup.

gyoshev:
Added some comments. Please assign reviewers when posting PRs, to keep each other up-to-date.

PR #91, opened 2017-01-11
feat(buttons): Add new design styles and improvements

theOrlin:


scss/button/_layout.scss
```diff
@@ -118,9 +118,21 @@
     }
 
     //DropDown Button
-    kendo-dropdownbutton.k-state-focused,
-    kendo-dropdownbutton:focus {
-        outline: 0;
+    .k-dropdownbutton {
```
joneff: If we have `k-split-button`, then this should be `k-dropdown-button`, or vice versa -- `k-splitbutton` and `k-dropdownbutton`.

The rest is fine.

joneff: @gyoshev @nlazarov  any thoughts on this?

joneff (changes requested):

scss/button/_theme.scss
```diff
@@ -1,9 +1,8 @@
 // TODO: button variables
-$button-focused-box-shadow: 0 0 0 2px darken($widget-bg, 6%) !default;
-$button-active-box-shadow: inset 0 3px 5px darken($widget-bg, 9%) !default;
+$button-focused-box-shadow: 0 3px 4px 0 rgba(0, 0, 0, .06) !default;
+$button-active-box-shadow: inset 0 2px 2px 0 rgba(0, 0, 0, .06) !default;
 
-$primary-focused-box-shadow: 0 0 0 2px lighten($primary-button-bg, 23%) !default;
-$primary-active-box-shadow: inset 0 3px 5px desaturate($primary-button-bg, 3%) !default;
+$primary-focused-box-shadow: 0 3px 4px 0 rgba(255, 99, 88, .06) !default;
```
gyoshev: How is this color calculated? seems related to `$accent`.

scss/button/_layout.scss
gyoshev: Seems to be the same question that we had for the `button-group` and `buttongroup`, no? telerik/kendo-angular-buttons#45

scss/button/_theme.scss
theOrlin: It is exactly the $accent color. I was unaware that I could use a hex variable inside an RGBA function in SASS. Corrected now.

scss/button/_layout.scss
theOrlin: Changed the class name to k-dropdown-button.

joneff (approved):

PR #99, opened 2017-01-12
fix: Adjust table wizard padding, sizing telerik/kendo#6508

theOrlin:


scss/editor/_layout.scss
```diff
@@ -181,6 +182,26 @@
...
+    .k-editor-table-wizard-dialog {
+        .k-content {
+            padding: $spacer-y $padding-x 0;
+        }
+
+        .k-edit-field {
+            .k-numerictextbox {
+                margin: 0 $padding-x-sm 0 0;
+
+                & + .k-dropdown {
```
joneff, line 192: `margin` should be on the following component, not the preceding one.

scss/editor/_layout.scss
```diff
@@ -142,6 +142,7 @@
 
 
     // Editor dialog
+
```
joneff, line 145: remove this space

scss/editor/_layout.scss
```diff
@@ -181,6 +182,26 @@
         }
     }
 
+    .k-editor-table-wizard-dialog {
+        .k-content {
+            padding: $spacer-y $padding-x 0;
```
joneff: `$spacer` should not be used for padding, rather only for margin. `$padding-x` is usually sufficient enough.

Not to mention that `$spacer` uses `em` and `$padding` uses `px`.

joneff (changes requested):

joneff (approved):

PR #119, opened 2017-01-16
Spreadsheet

joneff:


joneff:
Data validation dialog may look a bit odd due to a hidden button. I suppose after we handle telerik/kendo/issues/6043 everything will be OK.

PR #123, opened 2017-01-16
feat(mediaplayer): styles for mediaplayer

joneff:


scss/mediaplayer/_layout.scss
```diff
@@ -21,32 +21,41 @@
...
+
+        .k-toolbar-spacer {
+            margin-top: 0;
+            margin-bottom: 0;
+            border-width: 0;
+            flex: 1;
+        }
+
+        .k-align-right {
+            * + * { margin-left: $padding-x; }
```
gyoshev, line 56: Not really happy with the universal selectors. Are there no classes for these elements? If it is template content, it would be great to enclose these in a known element.

PR #126, opened 2017-01-16
feat: add pivotgrid styles

inikolova:


scss/pivotgrid.scss
```diff
@@ -0,0 +1,12 @@
+@import "variables";
+@import "mixins/all";
+
+
+// Dependencies
+@import "common/all";
+@import "dialog";
```
gyoshev, line 7: Does the field selector depend on the TreeView styles? If so, it should be listed, too.

scss/pivotgrid/_layout.scss
```diff
@@ -0,0 +1,176 @@
...
+    .k-pivot-filter-window .k-treeview {
+        max-height: 600px;
+    }
+
+    /* selector */
+    .k-fieldselector .k-edit-buttons {
+        bottom: auto;
+    }
+
+    .k-fieldselector .k-edit-label {
```
gyoshev: Consider nesting the `k-fieldselector` rules, it will make the file easier to read.

scss/pivotgrid/_theme.scss
```diff
@@ -0,0 +1,66 @@
...
+
+        &:hover,
+        &.k-state-hover {
+            @include appearance( hovered-button );
+        }
+
+        // Pressed state
+        &:active,
+        &.k-state-active {
+            @include fill( $button-text, $button-bg, $button-border, #{to top, $button-gradient} );
```
gyoshev, line 44: Does this match the pressed button state? If so, why no use `@include appearance( pressed-button );`?

inikolova: No, it does not match the pressed button state

scss/pivotgrid/_layout.scss
inikolova: Agree. Done!

scss/pivotgrid.scss
inikolova: Done!

PR #168, opened 2017-01-25
Add popup box shadow

inikolova:


scss/_variables.scss
```diff
@@ -134,7 +134,7 @@ $panel-border: rgba( black, .08 ) !default;
 $popup-text: $widget-text !default;
 $popup-bg: $base-bg !default;
 $popup-border: rgba( black, .08 ) !default;
-
+$popup-box-shadow: 0 2px 4px 0 rgba(0, 0, 0, .03), 0 4px 5px 0 rgba(0, 0, 0, .04) !default;
```
gyoshev: `$popup-shadow` is shorter, and doesn't omit information. The theme is not consistent in that regard, but now is a good time to start.

 Is the color different on purpose?

inikolova: Ok, I will change the variable name. As for the color, it has beed set after discussion with @yordanov

PR #171, opened 2017-01-27
Update dropdowns styling

inikolova:
Related to https://github.com/telerik/kendo-angular-dropdowns/issues/104

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -18,7 +18,6 @@
         display: inline-flex;
         vertical-align: middle;
         position: relative;
-        transition: background .2s ease;
 
         &[dir='rtl'] {
             text-align: right;
@@ -59,6 +58,7 @@
         border-style: solid;
         box-sizing: border-box;
         position: relative;
+        transition: all .1s ease; // sass-lint:disable-block no-transition-all
         // TODO: improve wrapper overflow
         //overflow: hidden;
         cursor: default;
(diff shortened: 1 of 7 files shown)


scss/common/_forms.scss
```diff
@@ -58,6 +58,7 @@
         border-style: solid;
         box-sizing: border-box;
         position: relative;
+        transition: all .1s ease; // sass-lint:disable-block no-transition-all
```
gyoshev, line 61: Do we really need to transition all properties? It seems like `transition: color .1s ease, background .1s ease;` will be sufficient.

scss/popup/_theme.scss
```diff
@@ -29,7 +29,10 @@ $popup-text: $normal-text-color !default;
 
     .k-list {}
 
-    .k-list .k-item {}
+    .k-list .k-item {
+        transition: all .2s ease; // sass-lint:disable-block no-transition-all
```
gyoshev, line 33: Ditto.

gyoshev:
Other than the minor transition question, the commits look great! :+1:

scss/common/_forms.scss
yordanov: `border-color` and `box-shadow` should also be subject to transition, as they change through different states too.

gyoshev: I see. Seems like `all` is okay for brevity.

PR #181, opened 2017-02-03
fix(dialog): do not modify layout of custom content

### LINKED ISSUE #180: Dialog content is not aligned properly
image

It seems that the dialog content is always aligned using `display: flex`, which breaks user content in it.

gyoshev:
closes #180


gyoshev:
Blocks the website deployment, so I'm merging this.

Issue #144, opened 2017-01-18
Missing styles for k-bare-button in toolbar


lkarakoleva:
**k-bare-button**
normal - color: inherit; opacity 0.7
hover - color: inherit; opacity 1
active - color: inherit; opacity 1
focused - background: currentColor; opacity 12% / 26% (accent)
![k-bare-button-states](https://cloud.githubusercontent.com/assets/16577622/22071090/bb3909ea-dda6-11e6-9f44-53544080d9ef.png)

joneff:
fixed by 8e1eb1313c9dc461d8c1b3a3a64c71cd216de846

PR #188, opened 2017-02-07
feat: bare button class

petyosi:
Working towards making "bare" buttons. So far, we can do it only in the Angular suite.

### DIFF
--- a/scss/button/_layout.scss
+++ b/scss/button/_layout.scss
@@ -224,7 +224,7 @@
     // Bare button
     .k-button.k-bare {
         // sass-lint:disable no-important
-        border-width: 0 !important;
+        border-color: transparent !important;
         color: inherit;
         background: none !important;
         box-shadow: none !important;
@@ -241,10 +241,10 @@
             pointer-events: none;
             position: absolute;
             z-index: 0;
-            top: 0;
-            left: 0;
-            right: 0;
-            bottom: 0;
+            top: -1px;
+            left: -1px;
+            right: -1px;
+            bottom: -1px;
         }
         &::after {
             content: "";
@@ -254,10 +254,10 @@
             pointer-events: none;
             position: absolute;
             z-index: 0;
-            top: 0;
-            left: 0;
-            right: 0;
-            bottom: 0;
+            top: 1px;
+            left: -1px;
+            right: -1px;
+            bottom: -1px;
         }
 
         // Hover state
--- a/scss/button/_layout.scss
+++ b/scss/button/_layout.scss
@@ -73,14 +73,6 @@
             box-sizing: content-box;
             justify-content: center;
         }
-        &-bare {
-            // sass-lint:disable no-important
-            border-width: 0 !important;
-            color: inherit;
-            background: none !important;
-            box-shadow: none !important;
-            // sass-lint:enable no-important
-        }
 
 
 
@@ -206,4 +198,27 @@
             border-bottom-right-radius: $border-radius;
         }
     }
+
+
+    // Flat button
+    .k-button.k-flat {
+        // sass-lint:disable no-important
+        border-width: 0 !important;
+        color: inherit;
+        background: none;
+        box-shadow: none;
+        // sass-lint:enable no-important
+    }
+
+
+    // Bare button
+    .k-button.k-bare {
+        // sass-lint:disable no-important
+        border-width: 0 !important;
+        color: inherit;
+        background: none !important;
+        box-shadow: none !important;
+        // sass-lint:enable no-important
+    }
+
 }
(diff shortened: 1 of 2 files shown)


gyoshev:
There is a ["k-button-bare" class](https://github.com/telerik/kendo-theme-default/blob/04693e308bf1e4ac5f086ce2f39bc5d2bfa6bfdd/scss/button/_layout.scss#L76-L83) that is used in several places: mediaplayer, colorpicker, pivotgrid, tabstrip, dialog, groupable. Can we reuse it?

I like the lack of `!important` in these styles, though. Maybe we could go without these, @joneff ?

petyosi:
I don't think that we can reuse it. The current approach keeps the button size consistent with the regular buttons. `k-button-bare` removes the border without compensating for it.

petyosi:
The current k-button-bare also does not accommodate for the design given.
bare-button-states
 @yordanov should call the shots here.

gyoshev:
I see. So the `.k-button-bare` uses only the icon color for its hover/active states, and the `.k-button.k-bare` uses backgrounds. If we need both types of button, maybe we should think of different names?

yordanov:
I was expecting that we would reuse this bare button in other components, hence the styles above. Here is how a bare button could look inside another component:

bare-buttons

In other words, the idea is the styles applied to `bare` button inherit default list item styles, not just removing background and border color of the `kendoButton`:
button-styles

I am expecting 99% of the time this button being used inside a wrapper as a part of a more complex UI, not just as a standalone button. To me that is the default use case for this type of button, and that is the only reason we introduced the `k-button-bare` class in ajax and k1.

Maybe there are other usage scenarios we are missing?

joneff:
Could we call this button "flat" and the one that changes just the color "bare"?

For reference: https://material.io/guidelines/components/buttons.html#buttons-flat-buttons

yordanov:
Makes sense. In this case, `bare` and `flat` will be identical in their normal state (provided that the font color is the same) and will have different interaction states:
 - `flat` inherits kendoButton styles
 - `bare` inherits its wrapper's font color and build the rest of the states based on this color.
 
If all agree, I guess, we'll need `primary` `flat` styles as well.

joneff:
@petyosi if you are OK with this change, I can go ahead and make the changes?

petyosi:
Let's see if we are all on the same page:

- we want to have `primary`, `flat` and `bare` modifiers of the buttons.
- we want those to map to `.k-button.k-flat`, `.k-button.k-bare`, and `.k-button.k-primary` CSS classes.
- we want to have `primary`, `bare`, and `flat` inputs for the Angular 2 components.
- question: what are we going to do with the `k-button-bare` instances?

If we all agree on it, let me know. @joneff can make the changes . Once done, I will add the `flat` input to the button component.

yordanov:
In addition:

- we want `flat` + `primary` and `bare` + `primary` play nice together
  - `.k-button.k-flat.k-primary` `.k-button.k-flat` placed together results in:
example
  - `.k-button.k-bare.k-primary` accepts the `primary` font color

joneff:
@petyosi, for the `k-button-bare` instances, I believe they are not so many, so as soon as we agree with the styles and add them not only for the k2 themes, but the classic as well, we can update them to match the new modifiers.

@yordanov your comment will be taken into consideration when updating the styles.

/cc @gyoshev @inikolova

petyosi:
Great, I guess this means I can transfer this pull request to @joneff to finalize.

joneff (changes requested):
As per the discussion, the styles will become `.k-button.k-flat`

yordanov:
Design issue: https://github.com/telerik/kendo/issues/6633

joneff:
While we are at it, should we add outline buttons? It's not like we have a use for them now, I know.

image

/cc @yordanov @petyosi @gyoshev

yordanov:
It's like we are trying to collect and support all UI languages and design patterns into one-size-fits-all theme. While it is not that bad to have these too, I would suggest we focus on and polish what we currently have in stock.

joneff (approved):

gyoshev:
Won't the [removal of the `k-button-bare` styles](https://github.com/telerik/kendo-theme-default/pull/188/files#diff-ae08dcf540fb1439bf514387c1208f52L76) break the Kendo UI for jQuery suite? Or is the rendering there already updated?

joneff:
after we merge this, I will go ahead and fix Kendo UI for jQuery suite including rendering, themes, wrappers, tests etc.

joneff:
@gyoshev @yordanov check the styles and if you don't find any issues, we can merge this.

gyoshev:
With the updated sizing that matches the other buttons, looks good to me. If @yordanov has no objections, :shipit:

yordanov (approved):
Go! Go! Go!

joneff:
related to #144

PR #209, opened 2017-02-16
fix: chart incorrect colors

inikolova:


tsvetomir:
Note that there's no opacity setting for the grid lines. We currently set opacity only on "area" series and series labels.

inikolova:
It appears rgba colors for major / minor GridLines are applies as exp
lines
ected:

tsvetomir:
You're right, opacity will be applied due to reading the colors as _rgba_.

tsvetomir (approved):

PR #227, opened 2017-02-20
Streamline window and dialog styles

joneff:


scss/_variables.scss
```diff
@@ -5,7 +5,7 @@ $font-family-monospace:  Consolas, "Ubuntu Mono", "Lucida Console", "Courier New
 
 $font-size-xs: 10px !default;
 $font-size-sm: 12px !default;
-$font-size-lg: 16px !default;
+$font-size-lg: 19px !default;
```
joneff, line 14: Used correct font size according to design

scss/_variables.scss
```diff
@@ -72,10 +72,10 @@ $cell-padding-y: $padding-y-sm !default;
...
 // $cell-padding-sm: $table-sm-cell-padding !default;
 
-$window-padding-x: $padding-x;
-$window-padding-y: $padding-y;
-$window-titlebar-padding-x: $padding-x;
-$window-titlebar-padding-y: $padding-x;
+$window-inner-padding-x: 16px !default;
+$window-inner-padding-y: 16px !default;
+$window-titlebar-padding-x: 16px !default;
+$window-titlebar-padding-y: 16px !default;
```
joneff, line 91: * renamed variables
* use correct values for variables

scss/common/_base.scss
```diff
@@ -36,6 +36,19 @@
...
+        width: 100%;
+        height: 100%;
+        background-color: #000000;
+        opacity: .5;
+        position: fixed;
+        top: 0;
+        left: 0;
+        z-index: 10001;
+    }
+
```
joneff, line 50: moved `.k-overlay` to `_base.scss`

scss/dialog.scss
```diff
@@ -5,8 +5,10 @@
 // Dependencies
 @import "common/all";
 @import "window";
+@import "button";
+@import "action-buttons";
```
joneff, line 9: improve component dependencies

scss/dialog/_theme.scss
```diff
@@ -1,108 +1,66 @@
...
 
-@include exports('dialog/theme') {
 
-    $window-header-text: $selected-text !default;
-    $window-header-bg: $accent !default;
-    $window-border: rgba(0, 0, 0, .08) !default;
-    $window-shadow-color: rgba(0, 0, 0, .3) !default; // remove the -color part
-    $window-shadow: 0 3px 3px 0 rgba(0, 0, 0, .06) !default;
-    $window-focused-box-shadow: 1px 1px 7px 1px $window-shadow-color !default;
+    .k-dialog {}
```
joneff: * remove variables
* inherit most of the styles from window

scss/window/_theme.scss
```diff
@@ -1,24 +1,37 @@
...
+
+    // Actions
+    .k-window-action {
+        opacity: 1;
+        transition: transform .1s ease-in-out;
+
+        &:hover,
+        &.k-state-hover {
+            opacity: 1;
+            transform: scale(2);
```
joneff: Not necessarily the best solution, but icons don't scale well  when not using factor of 2 e.g. 1.5

gyoshev: It's not ideal, to say the least. After a discussion with @yordanov, the close button seems to be a good candidate for a `.k-bare` button, without the scaling.

gyoshev:
Content layout in IE11 is broken due to the `.k-window-content { flex: 1; }`

image

I added `flex: 1 1 auto` in 9f59890, which resolves the problem in Angular. If it's good with the jQuery dialog, it could stay -- otherwise maybe flexbox is not the answer.

gyoshev (changes requested):
We should address the close button appearance without the scaling.

joneff:
@gyoshev I confirm. this solves the issue. 770af00 removes the scale-on-hover effect.

joneff:
@gyoshev even though 55e23d8 removes the styling, the action buttons still need `k-icon-only` class, otherwise dimensions will not be as accurate.

gyoshev (approved):

PR #229, opened 2017-02-20
Button size

joneff:
As discussed, buttons with icon only and other similar entities must be square for all themes.

For theme default, `.k-select` must be square.


magdalenaan:
Slider buttons are square but not with the right size: https://github.com/telerik/kendo-theme-default/blob/master/scss/slider/_layout.scss#L23-L24

Ideally, the slider buttons should be `.button-icon`

gyoshev (approved):

PR #230, opened 2017-02-20
fix: update grid layout to be usable by theme-bootstrap

joneff:
While vast, this refactoring is needed for Bootstrap theme to function properly.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -80,8 +80,8 @@ $nav-item-margin: 2px !default;
 $nav-item-padding-x: $padding-x-lg !default;
 $nav-item-padding-y: $padding-y-lg !default;
 
-$cell-padding-x: $padding-x-sm !default;
-$cell-padding-y: $padding-y-sm !default;
+$cell-padding-x: 12px !default;
+$cell-padding-y: 8px !default;
 $cell-padding: 8px !default;
 // $cell-padding-sm: $table-sm-cell-padding !default;
 
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -267,6 +267,8 @@
             line-height: $form-line-height;
             display: flex;
             align-items: center;
+            align-content: center;
+            justify-content: center;
             float: right;
             position: relative;
             z-index: 1;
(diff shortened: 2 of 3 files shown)


joneff:
Few screens to show the visual differences in theme-bootstrap and the lack of in theme-default

**theme-default**
before:
image

after:
image


**theme-bootstrap**
before:
image

after:
image

Don't mind the pager links in bootstrap theme. That's something entirely different

inikolova:
Things, which should be fixed: 

- Grid headers, toolbar, filter row heights - should be 46px, currently are the same as the grid cells (36px);
- Grid with aggregates and frozen columns - the whole layout is broken. Also, the layout in the Grid / Export to Excel demo;
- Grid grouping header must have the same height in both cases - grouping applied and without grouping;
- Missing top border on k-group-cell;
- Grid headers don't show ellipsis when text overflows;
- Double border on the last row - grid with frozen columns and aggregates;
- Wrong padding on grid filter / column menu popups;
- Wrong border colors of grid headers in IE11;

inikolova (changes requested):
Please see my comment

joneff:
Grid header, toolbar etc height fixed by f8622878093ff77d5acc765cde9b46831c123c61

joneff:
@inikolova I think I fixed all the points.

On a side note, since `flex` does not support `text-overflow: ellipsis` I switched to using `display: block` for same of the cases.

gyoshev:
@joneff please rebase and resolve the conflicts. Also, the commit messages won't push a new version -- consider squashing to a `feat:` or `fix:` commit.

inikolova:
After additional testing I observed the following: 

- _**k-grouping-header**_ is missing **display: block** definition;
- Currently **k-group-indicator** is 40px high while it should be the same as a regular button (30px); -> FIXED
- Grid **k-grouping-dropclue** is currently a vertical line, while it should have before / after elements; -> FIXED
- Wrong **.k-drag-clue** height -> FIXED

gyoshev:
@inikolova added the display:block. It is specific to the angular rendering, as we don't use block elements.

gyoshev (approved):

inikolova (approved):

PR #243, opened 2017-02-23
Fixes dropdowns

inikolova:


gyoshev (approved):

scss/popup/_layout.scss
```diff
@@ -71,7 +71,17 @@
     }
     .k-list {
         height: auto;
+
+        > .k-item {
+            display: flex;
```
joneff, line 76: can you add:

```scss
flex-direction: row;
align-items: center;
align-content: center;
```

joneff (changes requested):

scss/popup/_layout.scss
inikolova: OK, I added the last two declarations (the first one is a default value)

joneff (approved):

PR #259, opened 2017-03-01
chore: update treeview layout

joneff:
chore: move styles for k-image to icons.scss
chore: add styles for k-top mid bot; scope icon image sprite selectors

### DIFF
--- a/scss/treeview/_layout.scss
+++ b/scss/treeview/_layout.scss
@@ -21,6 +21,16 @@
             position: relative;
         }
 
+        // Wrappers
+        .k-top,
+        .k-mid,
+        .k-bot {
+            display: flex;
+            flex-direction: row;
+            align-items: center;
+            align-content: center;
+        }
+
         // Items
         .k-item {
             margin: 0;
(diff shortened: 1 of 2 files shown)


joneff:
purely cosmetic stuff. low priority

scss/treeview/_layout.scss
```diff
@@ -21,6 +21,16 @@
...
         }
 
+        // Wrappers
+        .k-top,
+        .k-mid,
+        .k-bot {
+            display: flex;
+            flex-direction: row;
+            align-items: center;
+            align-content: center;
```
gyoshev, line 31: It seems like the layout will change for the icons and checkboxes, when the item content wraps (has `white-space: normal`). Previously the icons and checkboxes were aligned with the item top, now they will be centered. Not sure if that is intentional.

joneff: Items are intentionally centered in the same way that we have set `white-space: nowrap`.

The change in layout comes from line 63 which adds right margin to the checkbox wrapper, which in terms is `1.5em` wide, which can be improved so it looks fine.

gyoshev (approved):

PR #260, opened 2017-03-01
fix(editors): common layout glitches

magdalenaan:


joneff (approved):

joneff:
There is only one caveat, about the line height in popups, but we'll deal with it in due time.

PR #262, opened 2017-03-02
fix(grid): filter row styling

inikolova:


scss/grid/_layout.scss
```diff
@@ -317,38 +317,50 @@
...
+        .k-widget,
+        .k-filtercell-wrapper > .k-textbox {
             display: flex;
             width: auto;
             flex: 1;
         }
+
+        .k-filtercell-wrapper > .k-textbox {
+            width: 100%;
+        }
```
gyoshev: Given the `flex: 1` above, why is the `width: 100%` necessary?

inikolova: There is default width **12.4em** applied to **.k-textbobox** and it should be overridden

gyoshev: There's a `width: auto` with the same selector on [line 350](https://github.com/telerik/kendo-theme-default/pull/262/files/80695152f353eefae8ba21dbe67ababd71051ade#diff-cee715f51ad80e6885203922176ce834R350)?

inikolova: The width must be 100% - I will modify the selector

joneff (approved):

PR #265, opened 2017-03-06
fix(panelbar): use correct variables for header padding

joneff:
Related to https://github.com/telerik/kendo-theme-bootstrap/issues/54

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -44,7 +44,7 @@ $popup-padding-x: 0 !default;
 $popup-padding-y: 0 !default;
 
 $header-padding-x: $padding-x !default;
-$header-padding-y: $padding-y !default;
+$header-padding-y: $padding-x !default;
 
 $form-line-height: $line-height !default;
 $form-line-height-em: $form-line-height * 1em !default;
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -44,7 +44,7 @@ $popup-padding-x: 0 !default;
 $popup-padding-y: 0 !default;
 
 $header-padding-x: $padding-x !default;
-$header-padding-y: $padding-x !default;
+$header-padding-y: $padding-y * 2 !default;
 
 $form-line-height: $line-height !default;
 $form-line-height-em: $form-line-height * 1em !default;
--- a/scss/panelbar/_layout.scss
+++ b/scss/panelbar/_layout.scss
@@ -16,7 +16,7 @@
             display: block;
 
             > .k-link {
-                padding: $header-padding-x;
+                padding: $header-padding-y $header-padding-x;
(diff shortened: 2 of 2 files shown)


inikolova (commented):
By design the padding on all sides should be **8px**:
https://github.com/telerik/kendo-ux/blob/master/themes/default/layout.psd

Maybe we could revert the change or add **padding: ($header-padding-y * 2) $header-padding-x;**

inikolova:
By design the padding on all sides is 8px -> check the following psd:
https://github.com/telerik/kendo-ux/blob/master/themes/default/layout.psd

joneff:
@inikolova I've updated `$header-padding-y` variable to be the same as `$header-padding-x` for theme-default.

inikolova:
This change will affect the mediaplayer titlebar - please review the mediaplayer widget, too

joneff:
It does not affect the media player title bar.

I did notice, however, that the media player toolbar uses different metrics and I will update those as well in additional pull request.

scss/_variables.scss
```diff
@@ -44,7 +44,7 @@ $popup-padding-x: 0 !default;
 $popup-padding-y: 0 !default;
 
 $header-padding-x: $padding-x !default;
-$header-padding-y: $padding-y !default;
+$header-padding-y: $padding-x !default;
```
gyoshev: I find this confusing in terms of variables -- changing `$padding-y` will have no effect on `$header-padding-y`, and changing `$padding-x` will. Why not `$header-padding-y: 2 * $padding-y;`? While the values are the same, it makes no sense to have a relationship between x and y variables.

inikolova: I agree with @gyoshev

joneff: That particular bit was discussed with @yordanov -- the padding is equal on all sides and is equal to to `$padding-x` i.e. changing the latter must and will change the padding in a predictable way.

gyoshev: I don't think it's predictable at all for customers to change vertical padding with `$padding-x`.

If the padding is always equal on all sides, then we don't need two variables. If it can be different (maybe in theme-bootstrap it is?), then we should maintain the relationships between `x` and `y` variables, in order for customization to be predictable.

yordanov: In terms of values, `x` and `y` must be equal for theme-default. In terms of variables, let's keep both `x` and `y` separate to allow easy and predictable customization.

joneff: @inikolova @gyoshev @yordanov So is that an approval or request for change?

gyoshev: @joneff it was a request for change

gyoshev (approved):

Issue #226, opened 2017-02-20
Displaced switch handle


yordanov:
http://www.telerik.com/kendo-angular-ui/components/inputs/switch/

![switch](https://cloud.githubusercontent.com/assets/1118895/23126368/001703d6-f77f-11e6-99ac-958b003f8eca.png)

PR #273, opened 2017-03-06
fix(switch): displaced switch handle

inikolova:
Closes telerik/kendo-theme-default#226


scss/switch/_layout.scss
```diff
@@ -53,7 +53,7 @@ $switch-handle-size: (.8 * $switch-size) !default;
         width: $switch-handle-size;
         height: $switch-handle-size;
         display: inline-block;
-        margin: 0 6px 0 3px;
+        margin: 0 ($padding-x-lg / 2) 0 ($padding-x-sm / 2);
```
joneff: The left margin should be the same as the top padding of `.k-switch-container`. Perhaps, we could extract it as a new variable e.g. `$switch-padding`.

On a side note, to me it's best to drop the margin altogether and use only the padding on `.k-switch-container`.

joneff (changes requested):
Only the left margin should be changed.

scss/switch/_layout.scss
inikolova: @joneff , you're right - I will make the needed changes. Regarding the padding - when long labels are used, the handle is moved. So I prefer keeping the margin

joneff (approved):

PR #277, opened 2017-03-07
fix(upload): move some metrics to the theme file

inikolova:


scss/upload/_theme.scss
```diff
@@ -132,4 +132,27 @@
     .k-dropzone-hovered {
         background-color: $hovered-bg;
     }
+
+    // Metrics
+    .k-dropzone {
+        .k-upload-status {
+            >.k-icon {
+                margin-right: $padding-x-lg / 2;
```
joneff, line 140: We have `$icon-spacing`, `$spacer` and `$spacer-x` variables that could be used.

joneff (approved):

PR #278, opened 2017-03-07
fix: split button in normal state misses borders, bg and border-radius

magdalenaan:
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/89


inikolova (commented):
There is outline applied to a focused splitbutton;

magdalenaan:
Native focus outline is removed. Focus and expanded states of split button are improved as well.

magdalenaan:
Active state of normal button should not be orange. It will be improved after fixing $button variables values.

joneff (approved):

PR #290, opened 2017-03-10
feat(buttongroup): add styles for responsive buttongroup

inikolova:
Related to telerik/kendo-angular-buttons#74

### DIFF
--- a/scss/button/_layout.scss
+++ b/scss/button/_layout.scss
@@ -177,7 +177,7 @@
 
     }
 
-    .k-responsive-button-group {
+    .k-button-group-stretched {
         display: flex;
 
         > .k-button {
--- a/scss/button/_layout.scss
+++ b/scss/button/_layout.scss
@@ -177,6 +177,21 @@
 
     }
 
+    .k-responsive-button-group {
+        display: flex;
+
+        > .k-button {
+            display: inline-block;
+            flex: 1 1 0;
+            overflow: hidden;
+            text-overflow: ellipsis;
+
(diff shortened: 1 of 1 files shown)


scss/button/_layout.scss
```diff
@@ -177,6 +177,21 @@
 
     }
 
+    .k-responsive-button-group {
```
gyoshev: In terms of classes, I would suggest using `.k-button-group-responsive` or `.k-button-group.k-responsive`. The latter is in sync with our decision for the `.k-button.k-bare` modifiers.
The name `responsive` does not describe what the styles do -- there are no responsive styles, after all. How about using `stretched` instead?

scss/button/_layout.scss
```diff
@@ -177,6 +177,21 @@
...
+        display: flex;
+
+        > .k-button {
+            display: inline-block;
+            flex: 1 1 0;
+            overflow: hidden;
+            text-overflow: ellipsis;
+
+            > .k-icon {
+                vertical-align: text-bottom;
```
gyoshev, line 190: Is this required only for this type of button group? It appears like something that should be applied to all button groups.

scss/button/_layout.scss
inikolova: Sounds great! I will check with @valchev 

Update: The class name has been changed to **k-button-group-stretched**:

https://github.com/telerik/kendo-angular-buttons/commit/dd31143

joneff (approved):

Issue #294, opened 2017-03-13
Edit row/cell - adjust the space between the input and the row/cell border


lkarakoleva:
According to the [design](https://github.com/telerik/kendo-ux/blob/master/themes/default/grid.psd) specification (grid, gantt, treelist), space between the input and the row/cell border should be decrease so text does not change position  in edit mode .

![untitled-1](https://cloud.githubusercontent.com/assets/16577622/23855577/ce3a8334-07fe-11e7-92e6-53120745894b.png)

PR #296, opened 2017-03-13
fix(grid): adjust the space between the input and row / cell border

inikolova:
Closes telerik/kendo-theme-default#294

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -448,6 +448,14 @@
         }
     }
 
+    .k-edit-cell,
+    .k-grid-edit-row > td {
+        >.k-textbox,
+        >.k-widget,
+        > .text-box {
+            margin-left: -($cell-padding + $input-border-width);
+        }
+    }
 
     // Resize handle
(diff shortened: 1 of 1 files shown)


scss/grid/_layout.scss
```diff
@@ -448,6 +448,14 @@
         }
     }
 
+    .k-edit-cell,
+    .k-grid-edit-row > td {
+        >.k-textbox,
+        >.k-widget,
+        > .text-box {
+            margin-left: -($cell-padding + $input-border-width);
```
joneff: this will yield a "mixed unit error" when used by bootstrap or any other theme not using pixels as a base unit.

joneff (changes requested):

scss/grid/_layout.scss
inikolova: @joneff , you're right. I will find another solution :)

inikolova: Using **calc** did the trick!

joneff (approved):

Issue #284, opened 2017-03-09
Grid headers misalignment


inikolova:
When a cell does not contail .k-link element, its content is moved to bottom. Can be observed in Grid data-binding demo ->  http://kendobuild-staging.openstack.progress.com/kendo-angular-ui/components/grid/data-binding/data-binding/

PR #297, opened 2017-03-13
Fixes/grid row height

joneff:
As discussed with @yordanov, we should keep hardcoded or calculated values for dimensions from layout, unless needed. Such is the case with certain dimensions in the grid, which force larger metrics in Bootstrap.

I've moved those into grid/theme.

Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/81

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -8,8 +8,6 @@
 
     $edit-cell-input-space: calc(-#{$cell-padding} - #{$input-border-width}) !default;
 
-    $delete-group-offset: calc(#{-$button-padding-y} - 1px) !default;
-
     .k-grid {
         display: block;
         position: relative;
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -174,18 +174,23 @@
 
         .k-link,
         .k-button.k-bare {
-            border-width: 0;
             padding: 0;
+            border-width: 0;
             display: inline-flex;
             align-items: center;
         }
 
-        .k-button.k-bare {
-            margin: $delete-group-offset auto $delete-group-offset $button-padding-x;
+        .k-link .k-icon {
+            margin-left: -( $icon-spacing / 2);
+            margin-right: $icon-spacing;
         }
 
-        .k-link .k-icon {
-            margin-right: $button-padding-y;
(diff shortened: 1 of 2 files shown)


inikolova:
In my opinion, dimensions should be kept in the **_layout** files. When overrides for different themes are needed, we could use variables. @joneff , @gyoshev  what do you think?

scss/grid/_layout.scss
```diff
@@ -254,8 +254,6 @@
 
 
         .k-header {
-            height: button-size();
-            line-height: button-size();
```
inikolova, line 260: This line-height is needed. Please check the following issue: 
https://github.com/telerik/kendo-theme-default/issues/284

inikolova (changes requested):

joneff:
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/81

scss/grid/_layout.scss
joneff: resolved in 5c96e00

joneff:
@inikolova I discussed the dimensions with @yordanov. It's easier to set constraints to a said place than to remove them. In terms, the layout is as generic as possible e.g. padding, border-styles etc, and we set restrictions in the theme files.

gyoshev:
@joneff can you elaborate why the heights are specific to the default theme and not the bootstrap theme? It seems like the filter row and grouping headers should be as high as the buttons and inputs, in any theme, and that the styles will be duplicated unnecessarily in the bootstrap theme. Maybe I'm missing something?

joneff:
@gyoshev What's specific for the default theme is that it's needed to have higher filter row and grouping headers than the standard row / header. That's not the case for bootstrap -- rows, headers, filter rows, toolbar etc all have equal height when having one line of content.

On a related note, definitely not the case in Material theme, which has smaller headers, but we'll get to that when we get to that.

gyoshev:
I confirmed with @yordanov that we don't need a fixed height of the rows for the default theme, as the row size should adapt to the content of the rows. Rebased and removed the sizing from the theme.

joneff:
I'm fine with the changes

inikolova (approved):

Issue #172, opened 2017-01-27
Wrong Button interaction states


yordanov:
All states must comply with the following guidelines:
https://github.com/telerik/kendo-ux/blob/master/components/buttons.psd

- [ ] :disabled, .k-state-disabled
- [ ] :normal, .k-state-default
- [ ] :hover, .k-state-hover
- [ ] :focus, .k-state-focused
- [ ] :hover:focus, .k-state-...
- [ ] :active, .k-state-active

theOrlin:
Those should be fixed now.
@yordanov

yordanov:
`:active` state is still wrong. It shoud have inner box-shadow only. The problem is that while in `:active` a `:focus` state is applied too, overriding the inner box-shadow of the `:active` state. We must apply the `:active` styles to the `:active:focus` state as well.

theOrlin:
Hopefully that does it.

@yordanov

yordanov:
- :focus state of button in a buttongroup must get inner shadow

Other states feel fine now.

@theOrlin @joneff @gyoshev @inikolova @lkarakoleva - What do you think about adding state transitions? Something like `transition: all .3s ease-in`? Are there any concerns regarding performance or behavior issues?

theOrlin:
When a transition is added with the current button styles, on click there is a slight delay between the mouse button release and the appearance of the focus state's shadow. This makes the buttons feel a bit sluggish, slow to respond.

joneff:
Related to #174

joneff:
@yordanov let's simplify states. And here is why:

When using a pointing device and an element is pressed, it's :hover :focus and :active at the same time. Actually, most of the time when an element is focused trough mouse interaction, it means is hovered. Or put simply, we don't really need a :hover:focus state, unless it differs from applying the box shadow of the focus onto the hover.

Then comes :active -- an element is active as long as the mouse is pressed onto it. And it does not apply to all components.

For instance, when is a combobox "active"? To me it's when it's expanded. What about a numerictextbox? Well it's actually never active. I mean it's focused, but the component itself is never active, rather the spin buttons get activated when clicked or the up / down arrows are pressed.

So "active" doesn't really mean anything.

Let's go with meaningful states.

And while we are on this page, let's pick a homogeneous naming -- we either have "focus" and "hover" or "focused" and "hovered". Having "selected" and "disabled" states, I'd go with the passive voice.

yordanov:
[...]

PR #300, opened 2017-03-14
Fixes/button variables

joneff:
As per discussion with @yordanov, some of the sates were not correctly used. The resulting changes in action buttons, calendar, colorpicker, multiselect and numerictextbox are synced and approved.

The changes in short: a default (gray) button should derive its pressed state from `$button-` variables family, while the selected / toggled state will be based on `$selected-` variables family.

Fixes https://github.com/telerik/kendo-theme-default/issues/172

### DIFF
--- a/scss/button/_theme.scss
+++ b/scss/button/_theme.scss
@@ -73,16 +73,16 @@
     // Button group
     .k-button-group {
 
-        // Active state
-        .k-button:active,
-        .k-button.k-state-active {
-            @include appearance( pressed-button );
-        }
-
         // Focused state
         .k-button:focus,
         .k-button.k-state-focused {
             @include appearance( focused-button );
+        }
+
+        // Active state
+        .k-button:active,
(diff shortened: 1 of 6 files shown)


scss/button/_theme.scss
```diff
@@ -76,13 +76,19 @@
         // Focused state
         .k-button:focus,
         .k-button.k-state-focused {
-            @include appearance( focused-button );
+            box-shadow: inset 0 0 0 2px rgba(0, 0, 0, .13);
```
inikolova: Can we extract a variable for this shadow?

joneff: I was thinking on that... However, we have two different sets of box shadow: one for "standalone" buttons and another for other types.

I was thinking perhaps a "generic" `$box-shadow` variable might do the trick, same as `$border-radius`...

inikolova (approved):

PR #307, opened 2017-03-14
fix(gantt): show delete-task-link only on task hover

joneff:
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/95
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/96


inikolova:
In my opinion, **k-task-draghandle** should be shown only on task hover. However, this might be a different commit

inikolova (approved):

joneff:
I am currently working on number of fixes for the gantt. I will include this fix as well.

joneff:
The fix for `k-task-draghandle` is in https://github.com/telerik/kendo-theme-default/pull/308

Issue #304, opened 2017-03-14
Gantt - update implementation according to the latest design 


lkarakoleva:
Update gantt according to the latest [design](https://github.com/telerik/kendo-ux/blob/master/themes/default/gantt.psd)

Issue #271, opened 2017-03-06
Gantt headers misalignment


inikolova:
Gantt headers in the timeline section are misaligned. I believe the issue is caused by the following commit: https://github.com/telerik/kendo-theme-default/commit/ed746b203a6b01041fe0a2953a49a38990f2b4f4

PR #308, opened 2017-03-14
Fixes/gantt fixes

joneff:
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/97
Fixes https://github.com/telerik/kendo-theme-default/issues/304
Fixes https://github.com/telerik/kendo-theme-default/issues/302
Fixes https://github.com/telerik/kendo-theme-default/issues/271

Note: somewhat related to https://github.com/telerik/kendo-theme-default/pull/297

### DIFF
--- a/scss/gantt/_layout.scss
+++ b/scss/gantt/_layout.scss
@@ -16,12 +16,6 @@
         .k-grid-header tr,
         .k-grid-content tr {
             height: calc( #{$row-height} );
-            line-height: inherit;
-
-            td,
-            th {
-                line-height: inherit;
-            }
         }
 
 
--- a/scss/gantt/_layout.scss
+++ b/scss/gantt/_layout.scss
@@ -196,7 +196,7 @@
     }
     .k-task-wrap {
         margin: 0 -1.5em;
-        padding: 0 25px;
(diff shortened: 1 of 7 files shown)


inikolova:
In my opinion, metrics should be kept in the _layout file

scss/gantt/_layout.scss
```diff
@@ -196,7 +196,7 @@
     }
     .k-task-wrap {
         margin: 0 -1.5em;
-        padding: 0 25px;
+        padding: 5px 25px;
```
inikolova, line 196: @joneff , should this padding be exactly 5 / 25px. Is it related to some metrics? Can we use variables?

joneff:
@inikolova  I agree. Still, in this particular case, we have different calculation of a grid header height. For instance in theme-default, grid header height is the sum of the button height plus the vertical cell padding, where as in theme-bootstrap it's as high as any other row.

inikolova (approved):

scss/gantt/_layout.scss
joneff: It's related to the size of `k-task-draghandle`. and thus far, we don't have such variable...

PR #323, opened 2017-03-17
fix(grid): expand/collapse icon is not centered

magdalenaan:
fixes https://github.com/telerik/kendo-theme-bootstrap/issues/82

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -77,7 +77,7 @@
         .k-group-col,
         .k-hierarchy-col {
             padding: 0;
-            width: 30px;
+            width: $icon-size*2;
         }
 
         .k-grouping-row p {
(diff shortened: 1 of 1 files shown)


inikolova (commented):
Can we change this hardcoded value 7px? Maybe we could use some calculations? For example (.k-group-col (which is 30px) - $icon-size (16px)) /2? What do you think?

scss/grid/_layout.scss
```diff
@@ -97,6 +97,11 @@
             text-overflow: none;
         }
 
+        .k-grouping-row .k-icon {
+            margin-left: addTwo(-1*$cell-padding-x, 7px);
```
joneff: change to 7px to `$icon-spacing / 2`

joneff (changes requested):

scss/grid/_layout.scss
magdalenaan: I think that there should be used $icon-size variable so icons could be centered in all cases.
If the header [k-group-cell width](https://www.screencast.com/t/TjCqs6jER) is ($icon-size*2), then "7px" can be replaced by ($icon-size/2).

joneff: Agree. You can go ahead with the changes.

joneff (approved):

PR #344, opened 2017-03-23
feat: add RTL styles for the Slider and Switch

tsvetomir:


gyoshev (approved):
The units are oddly specific, and should be taken into account for future refactoring.

tsvetomir:
Ouch, sorry! A PR with the proper units is on the way.

PR #352, opened 2017-03-27
fix(editor): make editor dialog content as wide as dialog

joneff:


scss/editor/_layout.scss
```diff
@@ -243,6 +247,78 @@
...
+        .k-filebrowser-toolbar {
+            margin: $spacer-y 0 0 0;
+            padding: $toolbar-padding-y $toolbar-padding-x;
+            display: flex;
+            justify-content: space-between;
+        }
+        .k-toolbar-wrap {
+            display: flex;
+
+            > * + * {
```
gyoshev, line 299: That seems to be a very intense selector. Can't we render a class instead?

magdalenaan (approved):

scss/editor/_layout.scss
joneff: Agreed. Will think about it

PR #361, opened 2017-03-31
fix(upload): different spacing in RTL mode

inikolova:
Closes telerik/kendo-angular-upload#67


gyoshev (approved):
TBH these can be extracted to a variable (or a list of all variables) so that they don't differ, whenever they are changed).

Issue #365, opened 2017-04-03
Loading indicator improvements


inikolova:
Currently, the loading indicator cannot change its size. Also, there are some backgrounds which must be the same as the wrapper background

cc: @yordanov , @gyoshev

PR #366, opened 2017-04-03
feat: loading indicator improvements

inikolova:
Closes telerik/kendo-theme-default#365

### DIFF
--- a/scss/common/_base.scss
+++ b/scss/common/_base.scss
@@ -418,51 +418,56 @@
         opacity: .3;
     }
 
-    $loading-stroke-width: 2px !default;
-    $loading-size: $icon-size - $loading-stroke-width !default;
-    $loading-speed: 2s !default;
-    $loading-color: $accent;
-    $loading-outer-size: ($loading-size + ($loading-stroke-width * 2));
-    $loading-half-size: ($loading-outer-size / 2) !default;
+    // Loading indicator
+    $loading-size: 1em !default;
+    $loading-base-font-size: 16 !default;
+    $loading-base-font-size-px: 16px !default;
+    $loading-font-size: 16 !default;
+    $loading-font-size-px: 16px !default;
+    $loading-stroke-width: calc( .03 *( #{$loading-font-size} / #{$loading-base-font-size}) * 1px) !default;
 
-    .k-icon.k-i-loading {
+    .k-i-loading {
         position: relative;
-        width: $loading-size;
-        height: $loading-size;
-        border: 2px solid $bg-color;
-        box-shadow: inset 0 0 0 $loading-stroke-width $loading-color;
-        border-radius: 50%;
-        box-sizing: content-box;
-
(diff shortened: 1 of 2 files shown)


scss/common/_base.scss
```diff
@@ -418,59 +418,63 @@
...
-        animation: k-loading-animation $loading-speed infinite ease;
+        margin-top: -($loading-size / 4);
+        margin-left: -($loading-size / 4);
+        width: $loading-size / 2;
+        height: $loading-size / 2;
+        animation: rotate reverse 1.4s linear infinite;
     }
 
-    @keyframes k-loading-animation {
+    @keyframes rotate  {
```
gyoshev: This may clash with animation names defined by the customer. Why not keep it `k-loading-animation`?

scss/common/_base.scss
```diff
@@ -418,59 +418,63 @@
...
-    $loading-outer-size: ($loading-size + ($loading-stroke-width * 2));
-    $loading-half-size: ($loading-outer-size / 2) !default;
-
-    .k-icon.k-i-loading {
+    // Loading indicator
+    $loading-size: 1em !default;
+    $loading-base-font-size: 16 !default;
+    $loading-base-font-size-px: 16px !default;
+    $loading-font-size: 16 !default;
+    $loading-font-size-px: 16px !default;
```
gyoshev: The default loading animation size should be `$icon-size`. If developers need to change the loading animation styles, they can do so via font-size on the element (like the case in the Grid > Export PDF).
It appears that we don't really need any of these variables.

scss/common/_base.scss
```diff
@@ -418,59 +418,63 @@
...
+    $loading-stroke-width: calc( .03 *( #{$loading-font-size} / #{$loading-base-font-size}) * 1px) !default;
```
gyoshev: As far as I recall, @yordanov wanted these to change based on the loading indicator size. If this is calculated in `em`s, it will follow the `font-size` set to the `k-i-loading` element.

gyoshev (changes requested):

scss/common/_base.scss
inikolova: You're right! My bad - I will fix this one

scss/common/_base.scss
inikolova: As far as I understood, the loading indicator as well as its border-width should depend on the font-size. @yordanov , could you please clarify?

scss/common/_base.scss
inikolova: There variables have been added because they are needed for the $loading-stroke-width calculations. However, I will check with @yordanov this comment, as well as the next one :)

scss/common/_base.scss
yordanov: loading

gyoshev (approved):

joneff (approved):

PR #388, opened 2017-04-12
chore(dropdowns): update dropdowns styling because of changed rendering

inikolova:
Related to:
https://github.com/telerik/kendo-angular-dropdowns/issues/201

https://github.com/telerik/kendo-angular-dropdowns/issues/202

### DIFF
--- a/scss/input/_theme.scss
+++ b/scss/input/_theme.scss
@@ -53,7 +53,7 @@
 
         .k-state-disabled & {
             &::selection {
-                @include appearance( "input" );
+                color: $input-text;
                 background-color: transparent;
             }
         }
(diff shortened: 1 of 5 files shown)


scss/input/_theme.scss
```diff
@@ -50,6 +50,13 @@
             background-color: $accent;
             color: $accent-contrast;
         }
+
+        .k-state-disabled & {
+            &::selection {
+                @include appearance( "input" );
+                background-color: transparent;
```
gyoshev: Seems more appropriate to use `color: $input-color` instead of `appearance( "input" )`, as the selection doesn't have a border and the background-color is overwritten

gyoshev (approved):

scss/input/_theme.scss
joneff: Agreed.

joneff (approved):

PR #391, opened 2017-04-13
feat: add styling for forms

inikolova:
Related to https://github.com/telerik/kendo-theme-default/pull/391

Closes: https://github.com/telerik/k2-site/issues/155

### DIFF
--- a/scss/input/_layout.scss
+++ b/scss/input/_layout.scss
@@ -5,8 +5,6 @@
     $textbox-shadow: 0 2px 2px 1px rgba(0, 0, 0, .06) !default;
     $fieldset-margin: 30px;
 
-    $input-border-width: 1px !default;
-
     .k-textbox,
     .k-input.k-textbox,
     .k-textarea {
@@ -85,7 +83,7 @@
     .k-checkbox-label,
     .k-radio-label {
         position: relative;
-        padding-left: $icon-size + ($padding-x-lg / 2);
+        padding-left: addTwo($icon-size, ($padding-x-lg / 2));
         vertical-align: text-top;
         cursor: pointer;
         display: inline-flex;
@@ -221,16 +219,16 @@
             &:last-child:last-of-type {
                 margin-bottom: 0;
             }
+        }
 
-            legend {
-                font-size: $font-size-sm;
-                font-weight: 600;
-                line-height: 1;
-                margin-bottom: $padding-y-lg * 2;
-                text-transform: uppercase;
-                padding: 0 $padding-x 0 0;
-                width: auto;
-            }
+        legend {
+            font-size: $font-size-sm;
+            font-weight: 600;
+            line-height: 1;
+            margin-bottom: $padding-y-lg * 2;
+            text-transform: uppercase;
+            padding: 0 $padding-x 0 0;
+            width: auto;
         }
 
         .k-form-field {
@@ -251,6 +249,12 @@
             margin-top: $padding-y * 2;
         }
 
+        .k-field-info {
+            display: inline-block;
+            font-size: ($font-size-xs / $font-size) * 1em;
+            margin: 0 $padding-x;
+        }
+
         .k-checkbox-label,
         .k-radio-label {
             margin-right: $padding-x * 2;
@@ -265,7 +269,7 @@
         }
 
         .k-form-field {
-            > span {
+            > span:not(.k-widget) {
                 display: block;
                 padding: $padding-y 0;
             }
@@ -274,44 +278,27 @@
                 width: 100%;
             }
         }
-
-        .k-field-info {
-            display: inline-block;
-            font-size: ($font-size-xs / $font-size) * 1em;
-            margin: 0 $padding-x;
-        }
     }
 
     .k-form-inline {
         $inline-form-element-width: 25% !default;
 
         .k-form-field {
-            > span,
-            > input {
-                display: inline-block;
-                box-sizing: border-box;
-            }
+            display: flex;
+            align-items: flex-start;
 
-            > span,
+            > span:not(.k-widget),
             > label:not(.k-checkbox-label):not(.k-radio-label) {
                 width: $inline-form-element-width;
-                float: left;
                 text-align: right;
                 line-height: $line-height;
-                padding: ($padding-y + $input-border-width) 0;
+                padding: addTwo($padding-y, $input-border-width) 0;
                 padding-right: $padding-x-lg;
-                box-sizing: border-box;
+                align-self: center;
             }
 
             > input {
                 width: calc(100% - #{$inline-form-element-width});
-                float: right;
-            }
-
-            &::after {
-                content: "";
-                display: block;
-                clear: both;
             }
         }
     }
--- a/scss/input/_layout.scss
+++ b/scss/input/_layout.scss
@@ -277,8 +277,8 @@
 
         .k-field-info {
             display: inline-block;
-            font-size: $font-size-xs;
-            margin-left: $padding-x;
+            font-size: ($font-size-xs / $font-size) * 1em;
+            margin: 0 $padding-x;
         }
     }
 
--- a/scss/input/_layout.scss
+++ b/scss/input/_layout.scss
@@ -5,13 +5,16 @@
     $textbox-shadow: 0 2px 2px 1px rgba(0, 0, 0, .06) !default;
     $fieldset-margin: 30px;
 
+    $input-border-width: 1px !default;
+
     .k-textbox,
-    .k-input.k-textbox {
+    .k-input.k-textbox,
+    .k-textarea {
         @include border-radius( $border-radius );
         padding: $input-padding-y $input-padding-x;
         width: 12.4em;
         box-sizing: border-box;
-        border-width: 1px;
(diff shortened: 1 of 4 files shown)


gyoshev (approved):

scss/input/_layout.scss
```diff
@@ -5,13 +5,16 @@
     $textbox-shadow: 0 2px 2px 1px rgba(0, 0, 0, .06) !default;
     $fieldset-margin: 30px;
 
+    $input-border-width: 1px !default;
+
     .k-textbox,
-    .k-input.k-textbox {
+    .k-input.k-textbox,
+    .k-textarea {
```
joneff: Do we need a separate class for textarea?

scss/input/_layout.scss
```diff
@@ -22,6 +25,11 @@
         -webkit-appearance: none;
     }
 
+    .k-textarea {
```
joneff, line 26: To continue the the above, this line could be

```scss
textarea.k-textbox {
    // styles go here
}
```

scss/input/_layout.scss
```diff
@@ -77,7 +85,7 @@
     .k-checkbox-label,
     .k-radio-label {
         position: relative;
-        padding-left: $spacer-x * 1.5;
+        padding-left: $icon-size + ($padding-x-lg / 2);
```
joneff: this will yield an error in theme-bootstrap

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+
+            &:first-child:first-of-type {
+                margin-top: 0;
+            }
+
+            &:last-child:last-of-type {
+                margin-bottom: 0;
+            }
+
+            legend {
```
joneff: the legend should always be in a fieldset, so we can "flatten" this selector

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+                font-weight: 600;
+                line-height: 1;
+                margin-bottom: $padding-y-lg * 2;
+                text-transform: uppercase;
+                padding: 0 $padding-x 0 0;
+                width: auto;
+            }
+        }
+
+        .k-form-field {
```
joneff: We can flatten this selector too

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+                margin-bottom: 0;
+            }
+        }
+
+        .k-alert-error {
+            font-size: $font-size-sm;
+            margin-top: $padding-y * 2;
+        }
+
+        .k-checkbox-label,
```
joneff: This could be merged with the above standalone styles

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+                display: block;
+                padding: $padding-y 0;
+            }
+
+            > input {
+                width: 100%;
+            }
+        }
+
+        .k-field-info {
```
joneff: what if I have inline form with field info?

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+                box-sizing: border-box;
+            }
+
+            > span,
+            > label:not(.k-checkbox-label):not(.k-radio-label) {
+                width: $inline-form-element-width;
+                float: left;
+                text-align: right;
+                line-height: $line-height;
+                padding: ($padding-y + $input-border-width) 0;
```
joneff: this will yield error in theme-bootstrap

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
         }
     }
 
+    // Forms
+    $form-spacer: $padding-x * 2 !default;
```
joneff, line 201: can we either use `$spacer`, which is meant for margins, or a new variable?

scss/input/_theme.scss
```diff
@@ -42,6 +43,10 @@
             background-color: $accent;
             color: $accent-contrast;
         }
+
+        &::placeholder {
+            color: rgba($text-color, .5);
```
joneff: this should be `rgba($input-text, .5)`

scss/input/_layout.scss
```diff
@@ -5,13 +5,16 @@
     $textbox-shadow: 0 2px 2px 1px rgba(0, 0, 0, .06) !default;
     $fieldset-margin: 30px;
 
+    $input-border-width: 1px !default;
```
joneff: we can extract this variable to _variables.scss

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+    // Forms
+    $form-spacer: $padding-x * 2 !default;
+
+    .k-form,
+    .k-form-inline {
+        font-size: $font-size;
+        line-height: $line-height;
+        padding: $form-spacer;
+
+        fieldset {
```
joneff, line 209: shouldn't `.k-form fieldset` and `.k-fieldset` have the same styles?

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+            display: inline-block;
+            font-size: ($font-size-xs / $font-size) * 1em;
+            margin: 0 $padding-x;
+        }
+    }
+
+    .k-form-inline {
+        $inline-form-element-width: 25% !default;
+
+        .k-form-field {
```
joneff, line 286: Are there any arguments against `display: flex`?

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+
+    .k-form {
+        input,
+        label:not(.k-checkbox-label):not(.k-radio-label),
+        .k-widget {
+            display: block;
+        }
+
+        .k-form-field {
+            > span {
```
joneff: What happens if I have a combobox as the input? It renders a span.

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+        }
+    }
+
+    .k-form-inline {
+        $inline-form-element-width: 25% !default;
+
+        .k-form-field {
+            > span,
+            > input {
+                display: inline-block;
```
joneff: do we really need this selector? Floated elements can have dimensions. The `box-sizing` property is defined in the next `span` selector, so it only needs to exist in the next `input` selector.

scss/input/_layout.scss
```diff
@@ -191,6 +199,122 @@
...
+                box-sizing: border-box;
+            }
+
+            > span,
```
joneff: Same question as with block forms -- what if my input is combobox? It renders a span.

scss/input/_layout.scss
inikolova: It appears that **.k-fieldset** has been previously defined, but not documented. **.k-form fieldset** follows the idea for not using additional classes for styling form elements

scss/input/_layout.scss
inikolova: $padding-x / $padding-y are used for margins in other places, too. I don't think we need additional variable here

scss/input/_layout.scss
inikolova: I think it is more meaningful to use **.k-textarea** for textarea elements. So I prefer keeping this class

scss/input/_layout.scss
inikolova: Same as the above

scss/input/_layout.scss
gyoshev: I could not find any references for `.k-fieldset` in the sources of Kendo UI for jQuery or Angular. Do we use it anywhere I'm missing or should we erase it?

scss/input/_layout.scss
inikolova: .k-alert-error has font-size defined too, so we cannot merge these rules

scss/input/_layout.scss
gyoshev: Good catch, it seems that we can use `span:not(.k-widget)`. Not too good performance, though :disappointed: ...

scss/input/_layout.scss
inikolova: I was not able to find any references to .k-fieldset, too.

scss/input/_layout.scss
inikolova: This class is specific to some sections in a form, so I prefer keeping the selector as it is

scss/input/_layout.scss
inikolova: Fixed

scss/input/_layout.scss
inikolova: No, I made the needed changes

scss/input/_layout.scss
inikolova: Fixed, as @gyoshev suggested

scss/input/_layout.scss
inikolova: This rule was modified as **display: flex** was added

scss/input/_layout.scss
inikolova: Fixed

scss/input/_layout.scss
inikolova: Fixed

Issue #261, opened 2017-03-01
Clear button overlaps the selected ComboBox item


siyka-marinova:
[https://www.screencast.com/t/XOEMXr58FDj](https://www.screencast.com/t/XOEMXr58FDj)
After selecting an item that is longer than the conbobox input size, its value overlaps the the clear button (after selection of an item, when the focus is on the combobox).

inikolova:
Related to: https://github.com/telerik/kendo-angular-dropdowns/issues/214

PR #395, opened 2017-04-18
fix(combobox): clear button overlaps the selected item

inikolova:
Closes telerik/kendo-theme-default#261

### DIFF
--- a/scss/combobox/_layout.scss
+++ b/scss/combobox/_layout.scss
@@ -20,6 +20,13 @@
                     right: auto;
                 }
             }
+
+            &.k-combobox-clearable {
+                .k-input {
+                    padding-left: $icon-size + $padding-x;
+                    padding-right: $input-padding-x;
+               }
+            }
         }
 
         // Compact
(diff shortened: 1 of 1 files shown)


gyoshev (commented):
It's a bit confusing to encounter the RTL styles first, and have the default styles at the end of the file. Why not use the same selectors in both cases (`&.k-combobox-clearable`)?

inikolova:
I added this rule at the end of the file in order to avoid unnecessarily increased specificity

gyoshev:
Valid point. The specificity for the RTL styles can also be reduced by moving them to the clearable selector like this:
```
.k-combobox-clearable {
  .k-input {
      padding-right: $icon-size + $padding-x;
  }

  &[dir='rtl'] .k-input {
     padding-left: $icon-size + $padding-x;
     padding-right: $input-padding-x;
  }
}
```
This way the declarations will also be kept together.

inikolova:
Isn't keeping all rtl styles in a single selector a better option?

gyoshev:
Usually, yes. But it is more confusing to have overwriting styles (like the RTL ones) above the styles they override:

    .foo[dir='rtl'] { float: left; }
    .foo { float: right; }

PR #400, opened 2017-04-20
feat(grid): set height on container

gyoshev:
This allows the grid height to be set on the container, and resizes the grid accordingly. Works also when the grid height not set. Should be merged in sync with the grid PR that adds `.k-grid-container` for NG2. Seems to be working fine with the jQuery component, too.


joneff (approved):

inikolova (approved):

sprell:
It looks like this change broke the jQuery grid when locked columns are in use. It will no longer display the unlocked columns correctly. I verified that it is working correctly in v2.30.0, but not in v2.31.0 (which looks to have only this change in it).

I have an example where I changed this jQuery grid demo...
http://demos.telerik.com/kendo-ui/grid/frozen-columns
...to use the new @progress/kendo-theme-default.

Here is the modified version...
http://dojo.telerik.com/aNUSa

All I did was replace the existing CSS with this...
http://www.telerik.com/kendo-angular-ui/npm/node_modules//@progress/kendo-theme-default/dist/all.css

gyoshev:
@sprell this is resolved in https://github.com/telerik/kendo-theme-default/commit/5d23b0932e12908c4e1e031721aa48f9c7ab0791 -- the latest Kendo UI for jQuery release includes the `k-display-block` class; until you upgrade, feel free to set the class to the grid:

    // 2017.1 and below
    <div id="grid" class="k-display-block"></div>

Here is the [updated Dojo](http://dojo.telerik.com/@gyoshev/EXoto) with the latest release.

PR #416, opened 2017-04-26
chore: add bootstrap compatibility check

gyoshev:
Adds a build step that verifies that the theme is compatible with variables in Bootstrap v4. Works well against mixed unit errors.

To achieve this, there is a `devDependency` on bootstrap v4. This should be kept in sync with the version of the kendo-theme-bootstrap dependency, preferably the latest Bootstrap release version.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,8 @@
 *.tgz
 .tmp/
 .travis.yml
-dist/*.js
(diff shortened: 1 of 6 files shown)


gyoshev:
Merging to fix kendo-theme-bootstrap build.

PR #437, opened 2017-05-01
fix(dialog): alignment in IE11 and Firefox RTL

gyoshev:
see telerik/kendo-angular-dialog#53


gyoshev:
Merging this as it affects only the dialog in a `.k-dialog-wrapper`, which is found only in the Dialog for Angular rendering. The fix solves critical regressions.

Issue #438, opened 2017-05-02
Pager buttons appear above modal dialog


inikolova:
Can be observed in this online demo: http://www.telerik.com/kendo-angular-ui/components/grid/editing/external-editing/

PR #442, opened 2017-05-02
fix(dialog): pager buttons appear above modal dialog

inikolova:
Fix for https://github.com/telerik/kendo-theme-default/issues/438


gyoshev (approved):
By the way, this also resolves issues with stacking order in IE :tada:

PR #443, opened 2017-05-02
fix(Grid): Remove negative margins from input elements in inline edit…

theOrlin:
… mode


joneff:
@gyoshev @inikolova @lkarakoleva this is a temporary solution for the edit-jump issue

joneff (approved):

PR #463, opened 2017-05-03
Fixes/spreadsheet toolbar

joneff:
Fixes https://github.com/telerik/kendo/issues/6549


scss/spreadsheet/_layout.scss
```diff
@@ -207,9 +207,15 @@
 
         // sass-lint:disable no-important
```
gyoshev, line 208: It is advisable to add a comment why the `!important` is unavoidable, for example like in [upload/_layout.scss](https://github.com/telerik/kendo-theme-default/blob/master/scss/upload/_layout.scss#L350)

magdalenaan (approved):

Issue #392, opened 2017-04-13
Inputs issues.


MilenaCh:
- [x] Invalid style of NumericTextBox is wrong. Arrorw keys should be colored in red.

 Design: https://www.screencast.com/t/q49KnecO
 Current: https://www.screencast.com/t/q49KnecO
- [x] Input's text is not centered in IE.
Actual: https://www.screencast.com/t/9wBei2Myu2

inikolova:
The second issue has been already addressed -> https://github.com/telerik/kendo-theme-default/issues/370

PR #464, opened 2017-05-03
fix(NumericTB): Fix invalid state spinners color

theOrlin:
Dealing with https://github.com/telerik/kendo-theme-default/issues/392

### DIFF
--- a/scss/numerictextbox/_theme.scss
+++ b/scss/numerictextbox/_theme.scss
@@ -42,6 +42,11 @@
     .k-numeric-wrap.k-state-invalid {
         border-color: $error;
         color: $error;
+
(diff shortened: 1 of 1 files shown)


scss/numerictextbox/_theme.scss
```diff
@@ -42,6 +42,11 @@
     .k-numeric-wrap.k-state-invalid {
         border-color: $error;
         color: $error;
+
+        .k-i-arrow-60-down,
+        .k-i-arrow-60-up {
+            color: $error;
```
joneff, line 48: We should probably use inherit in this particular case, but that's fine as well

joneff (approved):

PR #482, opened 2017-05-05
fix(Grid): Add validation tooltip styling

theOrlin:
Adding missing validation tooltip styling in the Grid described in https://github.com/telerik/kendo-theme-bootstrap/issues/218

### DIFF
--- a/scss/grid/_theme.scss
+++ b/scss/grid/_theme.scss
@@ -9,6 +9,9 @@
     $grid-grouping-row-text: $text-color !default;
     $grid-grouping-row-bg: darken($bg-color, 7%) !default;
 
+    $tooltip-color: $accent-contrast !default;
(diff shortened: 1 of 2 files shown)


scss/grid/_theme.scss
```diff
@@ -89,6 +92,28 @@
                 background-color: $grid-chrome-text;
             }
         }
+
+        .k-tooltip.k-tooltip-validation {
+            color: $tooltip-color;
+            background-color: $tooltip-bg;
+            @include border-radius($border-radius);
```
joneff: This must be in the layout (it already is, but not using `@include`)

joneff: This must be in the layout (it already is, but not using @include)

joneff (changes requested):

joneff (approved):

PR #489, opened 2017-05-10
Extend variables

joneff:
Fixes https://github.com/telerik/kendo-theme-bootstrap/issues/245

Add global options variables for box-shadow, border-radius, gradients and transitions

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1,3 +1,6 @@
+@import "mixins/colors";
+
+
 // Fonts
 
 /// Base font size across all components.
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1,6 +1,13 @@
 @import "mixins/colors";
 
 
+// Options
+$enable-rounded: true !default;
+$enable-shadows: true !default;
+$enable-gradients: true !default;
+$enable-transitions: true !default;
+
+
 // Fonts
 
 /// Base font size across all components.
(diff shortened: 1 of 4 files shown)


cornillemichiel:
I noticed the variables added in 254b611c03e7c590675c89b5aa1f6fa8503ed427 all have the _true_ default. In the bootstrap scss files shadows and gradients are set to _false_

joneff:
The way we've set up the themes to work is that theme-bootstrap (and any other theme) pulls mixins, but not variables, from theme-default. So we first need to make the mixins work for theme-default and then apply them to theme-bootstrap.

In terms, the variables in the said commit will apply only for theme-default. When this PR is closed, we'll add the same set of variables for theme-bootstrap which will respect values from bootstrap.

joneff:
@gyoshev, @inikolova, @mvgmvg  the question remains should we enforce `border-radius: 0` when the flag is false, or should we not apply border radius at all? The same goes for the rest of the flags.

@cornillemichiel it will be interesting to hear what you think about that question.

cornillemichiel:
Okay that makes sense thanks for the clarification.

cornillemichiel:
I'll get back to you on your second question

cornillemichiel:
@joneff I'd defo prefer no border radius at all. The less the framework gets in the way of customization the better. Also, less bytes to send over the wire 👍

inikolova:
@joneff, I prefer no border radius at all, too

joneff:
While we are still on the topic of `border-radius`, per say, should we go with a single value or add a per-component value, say `$button-border-radius: $border-radius !default`?

inikolova:
Using a single value sounds better to me, as all components have the same border-radius

inikolova (approved):

PR #497, opened 2017-05-15
feat(grid): Add styles for sorted column shading

theOrlin:
Dealing with https://github.com/telerik/kendo/issues/6906


scss/grid/_theme.scss
```diff
@@ -122,6 +122,10 @@
         }
     }
 
+    col.k-sorted,
+    th.k-sorted {
+        background-color: rgba( contrast( $panel-bg ), .02 );
```
joneff: it's an important variable, so it will be best if extract it as `$grid-sorted-bg`

joneff (changes requested):

joneff (approved):

PR #498, opened 2017-05-16
feat(timepicker): Add invalid state styles

theOrlin:
Dealing with https://github.com/telerik/kendo/issues/7129


joneff (approved):
I believe we could improve the invalid styles in terms, but this will do for now.

PR #499, opened 2017-05-16
feat(datetimepicker): Add invalid state styles

theOrlin:
Dealing with https://github.com/telerik/kendo/issues/7128

### DIFF
--- a/scss/datetime/_layout.scss
+++ b/scss/datetime/_layout.scss
@@ -53,13 +53,13 @@
     .k-datetimepicker {
 
         .k-i-warning {
-            right: $spacer * 4;
+            right: calc( #{ button-inner-size( 2 ) } + #{ $input-padding-x-sm });
         }
 
         .k-rtl &,
         &[dir='rtl'] {
             .k-i-warning {
-                left: $spacer * 4;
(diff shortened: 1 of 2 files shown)


scss/datetime/_layout.scss
```diff
@@ -30,12 +30,40 @@
...
                 right: auto;
             }
         }
     }
 
+    .k-datepicker,
+    .k-timepicker {
+
+        .k-i-warning {
+            right: $spacer * 2;
```
joneff, line 42: that's definitely not the correct position for the warning icon.

scss/datetime/_layout.scss
```diff
@@ -30,12 +30,40 @@
...
+            .k-i-warning {
+                left: $spacer * 2;
+            }
+        }
+    }
+
+    .k-datetimepicker {
+
+        .k-i-warning {
+            right: $spacer * 4;
```
joneff: neither is this

joneff (changes requested):
The icon position should be based around button-width and additional spacing. Something in the vicinity of `calc( button-size() + $input-padding-x)` or something similar.

Run it down with @lkarakoleva and @yordanov

joneff (approved):

Issue #493, opened 2017-05-12
Filterable DropDownList layout broken and default item not right-aligned in RTL mode


dtopalov:
In the following example:
[http://plnkr.co/edit/TuWFkSrUvArcpBdlCrPJ?p=preview](http://plnkr.co/edit/TuWFkSrUvArcpBdlCrPJ?p=preview)

Put any row in editing mode, and check the Discontinued column:

  1) Layout is broken if the DropDownList "filterable" option is set to true
  2) Default item is not right-aligned
  3) When the selected value does not pass the validation rules, ng-invalid is correctly applied, but there is no visual effect to indicate the input is invalid (same for the NumericTextBox and DatePicker)

1 and 2 seem to be resolved by the commented out styles in the example

PR #501, opened 2017-05-17
Dropdowns rtl issues and invalid state styling

inikolova:
Related to https://github.com/telerik/kendo-theme-default/issues/493

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -225,6 +225,20 @@
         right: 3px;
     }
 
+    .k-widget.ng-invalid {
+        .k-dateinput-wrap,
+        .k-dropdown-wrap,
+        .k-numeric-wrap,
+        .k-picker-wrap {
+            color: $error;
+            border-color: $invalid-border-color;
+
+            .k-select {
+                color: $error;
+            }
+        }
+    }
+
 }
 
 @include exports("forms/integration-with-bootstrap-forms") {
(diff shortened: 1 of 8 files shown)


gyoshev (changes requested):
Some things we discussed offline:

1. We should remove the rules that mention `&:invalid:not(.ng-valid)`, as these get applied in Kendo UI for jQuery by default, and do not look good
2. The invalid styling can be moved to `common/forms` and get cascaded through `.k-widget.ng-invalid`
3. The `$invalid-border-color` can perhaps be moved to `_variables.scss`.

inikolova:
The discussed changes have been made

gyoshev (approved):

scss/common/_forms.scss
```diff
@@ -225,6 +225,20 @@
...
+    .k-widget.ng-invalid {
+        .k-dateinput-wrap,
+        .k-dropdown-wrap,
+        .k-numeric-wrap,
+        .k-picker-wrap {
+            color: $error;
+            border-color: $invalid-border-color;
+
+            .k-select {
+                color: $error;
```
joneff: if the color is the same, shouldn't `color: inherit` do the same thing?

joneff (approved):

scss/common/_forms.scss
inikolova: @joneff , good catch! I will make this change

PR #512, opened 2017-05-23
Splitter - Corrected spelling to fix missing scrollbars

CodeBeastie:
The scrollbars were not appearing when using the Splitter with larger contents. Tracked it down to typo in the k-scrollable class name.


joneff (approved):

joneff:
Nice catch. Thanks

PR #519, opened 2017-05-30
fix(grid): Improve column selection checkbox position in RTL styling

theOrlin:
Dealing with https://github.com/telerik/kendo/issues/7138

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -196,7 +196,7 @@
                 }
 
                 .k-checkbox {
-                    margin-right: 7px;
+                    margin-right: calc(#{$padding-x-lg/ 2} + 1px);
                 }
             }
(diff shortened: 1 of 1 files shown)


joneff (approved):

scss/grid/_layout.scss
```diff
@@ -196,7 +196,7 @@
                 }
 
                 .k-checkbox {
-                    margin-right: 7px;
+                    margin-right: calc(#{$padding-x-lg/ 2} + 1px);
```
joneff, line 199: This seems like `$icon-size / 2`

joneff (changes requested):

joneff (approved):
Approving the request so we won't slow down the release, but we need to think it better...

Issue #537, opened 2017-06-12
Loading indicator is misaligned


inikolova:
Can be observed here: http://plnkr.co/edit/FW0Rocmc9bgUi495Gxwy?p=preview

inikolova:
Despite the change in the src, the loading icon is moved to the left with 2px

siyka-marinova:
Fixed in `@progress/kendo-theme-default v2.38.0`

PR #538, opened 2017-06-12
fix(combobox): misaligned loading indicator

inikolova:
Closes: https://github.com/telerik/kendo-theme-default/issues/537

### DIFF
--- a/scss/combobox/_layout.scss
+++ b/scss/combobox/_layout.scss
@@ -5,7 +5,8 @@
             > .k-i-loading {
                 top: 50%;
                 transform: translateY(-50%);
-                margin-right: $padding-x;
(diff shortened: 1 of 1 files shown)


gyoshev:
Seems like the styles position the loading icon correctly, but maybe the rendering in Angular is different from the one in Kendo UI for jQuery? See [this plunker](http://plnkr.co/edit/g271HrenyrVmllgpSiPt?p=preview).

joneff:
@gyoshev  In the KendoUI for jQuery, loading icon replaces open icon hence it's in the correct position.

gyoshev:
In that case, it may be more appropriate to fix this in the component, so that the markup is the same.

PR #542, opened 2017-06-13
fix(numerictextbox): spinner icons alignment; fix select width

inikolova:
Related to: https://github.com/telerik/kendo-theme-bootstrap/issues/258


joneff (approved):
seems legit

PR #543, opened 2017-06-16
fix(grid): Remove font icon in custom Edit/Delete buttons

theOrlin:
Dealing with https://github.com/telerik/kendo-ui-core/issues/2715


joneff:
I reject the commit. We should handle things within the code of the grid

PR #545, opened 2017-06-16
fix(dropdowns): misaligned loading indicator; wrong select width

inikolova:
Related to https://github.com/telerik/kendo-theme-default/issues/537

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -141,9 +141,16 @@
         .k-clear-value,
         .k-i-loading {
             position: absolute;
-            right: $padding-x;
             top: 50%;
             margin-top: -.5em;
+        }
+
+        .k-clear-value {
+            right: $padding-x;
+        }
+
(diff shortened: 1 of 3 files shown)


scss/common/_forms.scss
```diff
@@ -141,12 +141,19 @@
         .k-clear-value,
         .k-i-loading {
             position: absolute;
-            right: $padding-x;
             top: 50%;
             margin-top: -.5em;
         }
 
         .k-clear-value {
+            right: $padding-x;
```
joneff, line 149: shouldn't we combine this selector and the one below?

joneff (changes requested):

scss/common/_forms.scss
inikolova: @joneff , you're right -> thanks for the tip

joneff (approved):

Issue #547, opened 2017-06-16
Add the ability to change the disabled style mixin


IanMuirhead:
### Issue
Currently the `/scss/mixins/_disabled.scss` doesn't allow for the developer to modify the look and feel of the disabled state for buttons and other components.

### Proposed fix
Extend the mixin to include variables.

PR #548, opened 2017-06-16
feat: add variables to disabled mixin

IanMuirhead:
#547 

Adds variables for the disabled mixin to allow for styling customization of the disabled state of components

Includes:
- opacity
- grayscale

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -389,3 +389,7 @@ $chart-area-opacity: .8 !default;
 $icon-size: 16px;
 $icon-size-lg: 32px;
 $icon-spacing: 4px !default;
+
+// Disabled mixin variables
+$disabled-filter: grayscale(.1) !default;
+$disabled-opacity: .6 !default;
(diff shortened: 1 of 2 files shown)


IanMuirhead:
@joneff Anyone able to put some thought into this pull request?

gyoshev:
@IanMuirhead thank you for the contribution! The values can be handy when customizing the theme.

It seems that defining the complete filter (rather than the grayscale component) might be more flexible:

    $disabled-filter: grayscale(.1);

    filter: $disabled-filter;

This way, the filter function can also be changed (or completely disabled) with the variable. What do you think?

joneff:
We could, perhaps, use a mixin for this and if the argument is not specified e.g. `null` not to output any styles at all.

IanMuirhead:
@gyoshev Fair point - will make some modifications and update the pull request.

@joneff I had originally thought about using a mixin. Following the patterns elsewhere in the theme I thought variables would be more consistent.

joneff (approved):

gyoshev:
@IanMuirhead thanks for the update! It has been merged into the development channel, and will be included in the next release of the theme. :tada:

Issue #571, opened 2017-06-29
Extending dialog /window to include variables for customization 


IanMuirhead:
I'm trying to add variables to the `dialog` component and properly import it into the theme.

From what I can tell the theme isn't being included properly and I am unsure how to troubleshoot the issue.

I have a forked repository with a branch with the changes I've made so far. 

https://github.com/copperleaftech/kendo-theme-default/tree/feat/dialog-variables

An example of what is confusing me. You can see that the classes that are being added are `k-window-title k-dialog-title` but in the generated CSS there is no declarations for `k-dialog-title`.

![image](https://user-images.githubusercontent.com/884395/27666091-462ff22c-5c26-11e7-8294-79691b0fc595.png)

If you look at the scss files - I am perplexed as to why it isn't being generated. I have included a snippet of the declaration I've added to my fork below.

```   
 .k-dialog-title {
        color: $dialog-title-font-color;
        margin: 0;
        font-size: $dialog-title-font-size;
        line-height: 1;
        text-overflow: ellipsis;
        overflow: hidden;
        cursor: default;
        flex: 1;
    }
```

I think I'm missing something really obvious here - any idea if you can nudge me in the right direction?

gyoshev:
By design, the dialog inherits styles from the window component, which is [available in Kendo UI for jQuery](http://demos.telerik.com/kendo-ui/window/index). The dialog component extends these styles to add the action buttons functionality. Most likely the introduced variables need to work with the window styles.

IanMuirhead:
@gyoshev Understood.  

Looks like I had something irked on my local not pulling in my virtual folders properly. d'oh.

I'll fix up my implementation and submit a pull request to see if you're interested in including it in the main theme.

Issue #554, opened 2017-06-20
TabStrip colors


YKdvd:
The [tabstrip/_theme.scss file defines some colors at the top of the file](https://github.com/telerik/kendo-theme-default/blob/2241e16303659e4013693828ac6ad8550d3e6ae1/scss/tabstrip/_theme.scss#L1-L6):
```scss
$tabstrip-tabs-color: $accent !default;
$tabstrip-border-color: rgba(0, 0, 0, .08) !default;
$tabstrip-hover-color: blend-multiply($accent, darken($widget-bg, 8%)) !default;
$tabstrip-active-color: $base-text !default;
$tabstrip-active-background-color: desaturate($widget-bg, 1%) !default;
$tabstrip-disabled-color: darken($widget-bg, 27%) !default;
```
However, these never seem to be used anywhere.  Later on, color is set up as in:
```scss
    .k-tabstrip {
        .k-item {
            color: $link-text;
```
"$link-text" is indirectly defined as "$accent" elsewhere, but wouldn't it make sense to have
```scss
color: $tabstrip-tabs-color;
```
directly - that would seem to be the purpose these were defined for?.  That way, anyone overriding can simply do something like:
```scss
$tab-strip-color: [whatever color is desired];
/* other tabs- color variables overriden also */
@import "kendo-theme-default/scss/all"
```
instead of having to override all the tab stylings that define color and keep synced with any changes in the future?

In our case, we want to change the tabs to use the "selected" styling.  It would be helpful if a "$tabstrip-background-color" variable were defined (it could be "inherit"), and explicitly used to define "background-color: $tabstrip-background-color;" in k-item, so that could be overridden as well.

IanMuirhead:
@inikolova 

I am currently trumping these in our custom theme as well - I also set a custom `:hover`, `:focus` and `.k-state-active` 

We also change a number of other attributes, such as: `border-left`, `border-right`, `border-top`, `border-bottom`, `margin-bottom` ...

Our theme ends up skinning the kendo theme to look like the following: 
![tabstrip-example](https://user-images.githubusercontent.com/884395/27393208-85a8094a-565d-11e7-839a-fe0bb352efe0.png)

Related specifically to the color settings on the tabstrip, our use case could be partially covered by the following as well.

`$tabstrip-tab-color`
`$tabstrip-tab-color-hover-focus`
`$tabstrip-tab-color-active`

```
.k-tabstrip > .k-tabstrip-items .k-item {
[...]

PR #574, opened 2017-06-29
Feat/dialog variables

IanMuirhead:
#571 

# Purpose
To add the ability to theme both dialog and window components

# Details
Created a set of variables for both window and dialog
By default the dialog inherits all of the variables from the window theme
Expanded the `dialog/_theme` and `dialog/_layout` to have a matching set of variables which will allow for the users to create theme for either component


IanMuirhead:
I can squash these commits once reviewed and is seen as something you want to bring into the theme.

scss/_variables.scss
```diff
@@ -393,4 +388,46 @@ $icon-spacing: 4px !default;
...
\ No newline at end of file
+$disabled-opacity: .6 !default;
+
+// Window
+$window-inner-padding-x: 16px !default;
+$window-inner-padding-y: 16px !default;
+$window-titlebar-padding-left: 16px !default;
+$window-titlebar-padding-right: 16px !default;
+$window-titlebar-padding-top: 16px !default;
+$window-titlebar-padding-bottom: 16px !default;
```
gyoshev: I suggest following the `-x` / `-y` naming convention, or exposing a single variable (`$window-titlebar-padding`) that allows styling through the padding short-hand.

scss/_variables.scss
```diff
@@ -393,4 +388,46 @@ $icon-spacing: 4px !default;
...
+$window-titlebar-border-width: 0 0 1px !default;
+$window-titlebar-border-style: solid !default;
+
+$window-titlebar-min-height: 16px !default;
+
+$window-title-text-transform: inherit !default;
+$window-title-font-size: $font-size-lg !default;
+$window-title-font-family: inherit !default;
+
+// Dialog - by default inherits from window
```
gyoshev: Up until now, the theme has maintained consistent looks between the window and dialog components, this is why there have been no `$dialog-` variables. I suggest removing them and using the `$window-` variables instead, so that the consistency is maintained.

scss/_variables.scss
```diff
@@ -393,4 +388,46 @@ $icon-spacing: 4px !default;
...
+$window-inner-padding-x: 16px !default;
+$window-inner-padding-y: 16px !default;
+$window-titlebar-padding-left: 16px !default;
+$window-titlebar-padding-right: 16px !default;
+$window-titlebar-padding-top: 16px !default;
+$window-titlebar-padding-bottom: 16px !default;
+
+$window-titlebar-background-image: none !default;
+$window-titlebar-background-color: $accent !default;
+$window-titlebar-color: $accent-contrast !default;
```
gyoshev: This is a nitpick, but `background` is shortened to `bg` in all variable names, and text color uses the suffix `-text` rather than `-color`.

Also, what do you think about using only one variable that sets the background shorthand value? This will allow users to set the `background-repeat` and `background-position`, too.

gyoshev (commented):
Thank you for the contribution! The provided variables will indeed add a lot of flexibility to the component. I am adding some suggestions about how to make them easier to use. The common theme in the comments is that we strive to reduce the amount of variables to the bare minimum that allows enough flexibility, but does not overwhelm. Please take a look and let me know what you think.

IanMuirhead:
Thanks for the review!

Do you have your shorthand / other code styles documented anywhere? Would
be a great help.

I think all your suggestions make alot of sense - I'll submit a new pull
request tomorrow.

gyoshev:
I am afraid that we lack code style documentation at the moment, and we rely mostly on the context. I realize that providing such a document will be useful for future contributions, so I'll log it.

By the way, if you push commits to the same branch, this PR will update automatically, so no need for a new one.

IanMuirhead:
@gyoshev Should have all the changes requested included.

I went with the short-hand option for the padding, as in my use-case I need that granularity of control.

joneff:
@IanMuirhead we want to provide the minimum possible set of variables that's sufficient for common customization and thus avoid configuration madness.

It's not like we can't have variables for each and every aspect of a theme. Quite the contrary. We could, but then the theme would be hard as hell to customize. That being said, we've deliberately avoided variables like `*-text-transform` or `*-font-family`. Even those extra two variables would bring enough overhead if applied to all components.

For instance, we do not set font-family because we want it to be inherited from the page, or changed via simple rule `.k-widget {font-family: "Monaco", sans-serif}`.

Then there is the question of theme guidance -- our designers have designed the theme with certain things in mind. They build a visual language, a guide, in the same way that Google have built Material. In this particular case, they've designed the dialog and window to have different titlebar background colors, with the dialog having a more prominent one, which in terms is tied to the main theme color.

When we implement a theme, we take that into consideration and make the theme that way. That's why at times we make deliberate choices to allow or disallow a certain customization so the theme looks consistent. Point in case, setting text transform in the titlebar of window / dialog.

To elaborate even further, take a look at [themebuilder for UI for Asp.Net Ajax](http://themebuilder.telerik.com/aspnet-ajax) and more specifically, the Material skin. Like I've mentioned above, Google provides certain guide lines and using those guide lines, we've open for customization just three variables: primary color, accent color and base color. Should someone want to deviate from the guidelines, he or she must add explicit styles for that customization.

Lastly, but definitely not least, a theme that allows a much richer set of customization makes sense. Even the fact that you need the said customization hints the need for such a theme.

I will further review your changes and see how we can integrate them the same way #554 was integrated with #560.

I will follow up with another comment on variable naming, variable values etc.

joneff:
**Variable naming**

This one is fairly simple:

* we try variables to be as descriptive as possible, possibly containing the component name;
* we also shorted background to bg and droped the color suffix for colors;
* we use directional variables for x and y;
* we could put -text after -bg, because we could use color functions;
* we could put -border last for the same reason;
* we have interactive states for `:hover` and `:active`, but those are a bit tricky...

or to sum up, say we want variables for a button:

```scss
$button-padding-x: 4px;
$button-padding-y: 10px;
$button-border-width: 1px;
$button-bg: #880000
$button-text: contrast-wcag( $button-bg );
$button-border: darken( $button-bg, 10 );
$button-gradient: null;
```

Since we don't really need variables for button margin or border style (presumably, if a border is present, it's solid), we leave those out.

**Introducing new variables**

We try to introduce a new variables as little as possible and reuse existing variables as much as possible to ensure consistency, but allow for customization.

As a general rule, variables that apply to more components are placed in [_variables.scss](https://github.com/telerik/kendo-theme-default/blob/develop/scss/_variables.scss) and more specific variables are placed in their respective theme files. 

So say we want to style the window and the dialog, that would mean putting the variables `$window-titlebar-bg/color/border` and `$dialog-titlebar-bg/color/border` in [window/_theme.scss](https://github.com/telerik/kendo-theme-default/blob/develop/scss/window/_theme.scss) and [dialog/_theme.scss](https://github.com/telerik/kendo-theme-default/blob/develop/scss/dialog/_theme.scss) repsectively.

**Mixins**

To avoid mistakes and enforce consistency, we use mixins a lot. A big part of those are the `appearance($what)` mixins and `fill(...)` mixin.

For instance, `appearance(button)` will use use the relevant `$button-` variables and not only that, it will combine all selectors including it, by using placeholder selectors internally. The idea behind is to have the slimmest possible output.

The same applies for header, panel, input etc...

In the case of the window / dialog, since we'll be overriding the appearance, we could use `fill(..)` mixin.

**Variable linking / reusing**

Following up on my previous comment, that want to reuse variables and still allow customization, we want to set the default value of the newly created variables to value prior the introduction. Looking at the code in window/_theme.scss, we see `appearance(header)` hence we know we want to use `$header-*` variables:

```scss
// window/_theme.scss
$window-titlebar-bg: $header-bg;
$window-titlebar-text: $header-text;
$window-titlebar-border: inherit;

// dialog/_theme.scss
$dialog-titlebar-bg: $accent;
$dialog-titlebar-text: contrast-wcag($dialog-titlebar-bg);
$dialog-titlebar-border: inherit;
```

Notice how we link `$dialog-titlebar-text` to the `$dialog-titlebar-bg` and not `$accent-contrast` as it was originally. We want to keep the text to background contrast ratio, hence, we link both variables.

**Compatibility with Bootstrap and other frameworks**

That part is also important. We want to be compatible and consistent with Bootstrap (and other prominent frameworks) as much as possible and that's why we try to keep our variables close to theirs, sometimes using event the same name for some variables, say `$input-bg`.

Since Bootstrap mainly has `$*padding-x` and `$*padding-y` variables we decided to go that way. Yes, Bootstrap uses both x/y padding and shorthand values, but we want to be consistent, so we only go with x/y padding. That tends to help in certain scenarios, when you want to place something from the left / right edge. Sure we could use `first($list)` / `last($list)` functions, but having explicit values does not hurt.

scss/_variables.scss
```diff
@@ -97,11 +97,6 @@ $cell-padding-y: 8px !default;
 $cell-padding: 8px !default;
 // $cell-padding-sm: $table-sm-cell-padding !default;
 
-$window-inner-padding-x: 16px !default;
-$window-inner-padding-y: 16px !default;
-$window-titlebar-padding-x: 16px !default;
-$window-titlebar-padding-y: 16px !default;
```
joneff: Those variables are good here.  We first keep "layout" or "metrics" variables and then we put the "theme" or "skin" variables.

scss/dialog.scss
```diff
@@ -8,5 +8,5 @@
 
 
 // Component
+@import "dialog/layout";
```
joneff: good spotting those two are inverted.

scss/dialog/_theme.scss
```diff
@@ -5,15 +5,10 @@
 
 
     // Titlebar
-    .k-dialog-titlebar {
```
joneff: The dialog is different from the window by design. However, we could introduce dialog variables in this file.

scss/window/_layout.scss
```diff
@@ -21,19 +21,23 @@
 
     // Title bar
     .k-window-titlebar {
-        padding: $window-titlebar-padding-y $window-titlebar-padding-x;
-        min-height: 16px;
-        border-width: 0 0 1px;
-        border-style: solid;
```
joneff: The min height is actually as big as an icon (the minimum content). Actually, I am not sure we need it since we use flexbox... We could definitely improve this bit.

Border is 1px solid by design e.g. in order to fit in the theme we must have it like this. We could "hide" it by setting `$window-titlebar-border: transparent`.

scss/window/_layout.scss
```diff
@@ -21,19 +21,23 @@
...
         align-items: center;
     }
+
+    // Title bar title text
     .k-window-title {
         margin: 0;
-        font-size: $font-size-lg;
+        font-size: $window-title-font-size;
+        font-family: $window-title-font-family;
+        text-transform: $window-title-text-transform;
```
joneff: I am still wondering for these. Font size make sense; fonts are inherited, by design and that would be the first text-transform variable we have...

scss/window/_theme.scss
```diff
@@ -18,14 +18,17 @@
     // Titlebar
     .k-window-titlebar {
         @include appearance( header );
+        color: $window-titlebar-text;
         border-color: inherit;
+        background: $window-titlebar-bg;
```
joneff: Since we are overriding the appearance, we don't really need `appearance(header)`. In terms, this could become:

```scss
.k-window-titlebar {
    @include fill( $window-titlebar-text, $window-titlebar-bg, $window-titlebar-border);
}
// or, the longer form
.k-window-titlebar {
    border-color: $window-titlebar-border;
    color: $window-titlebar-text;
    background-color: $window-titlebar-bg;
}

**Review by joneff — CHANGES_REQUESTED** [review]

File: scss/window/_layout.scss (hunk 8 above)
Comment (IanMuirhead) [reply scss/window/_layout.scss]: Our specific design system calls for all titles, h1 - h6 and dialog / windows to be a specific font-family and in dialog / modals for them to be uppercased.

File: scss/_variables.scss (hunk 4 above)
Comment (IanMuirhead) [reply scss/_variables.scss]: Understood - will bump them back up.

File: scss/dialog/_theme.scss (hunk 6 above)
Comment (IanMuirhead) [reply scss/dialog/_theme.scss]: Ahh - I was misinformed then.  

Will correct and introduce some new variables.

File: scss/window/_layout.scss (hunk 7 above)
Comment (IanMuirhead) [reply scss/window/_layout.scss]: I was wondering the need for the min-height as well - it wouldn't be required.

As per the borders - I haven't removed the default styling but rather just moved them into variables so that they could be customized if required.

To explain the reasoning behind the shorthand padding I have included a screenshot of the outcome of these variables.

The close button is being hovered.

copperleaf-headerbar-design

**Comment by IanMuirhead** [comment]
@joneff Thanks for the review and expansive reply about the decisions going into the design of the theme.

At Copperleaf we are also in the midst of building our own design system, and a key component of that system will be the Kendo Angular Framework. 

For our design system to be able to adapt within the constraints of the kendo themes we have two options: 

- I fork the theme repository and pull in changes as you continue to maintain and I do not contribute to the product
- I am able to contribute pieces of the customization we require for our specific use case into the main theme

I do understand your desire to keep the theme as simple as possible - but I don't think our use cases will be unique. My preference would always be bringing changes which will allow me to theme Kendo without the need to have trumps for certain styles. That will allow my team to more effectively use the theme.

I am going to update my PR with some of your review changes.  I can respect your choice to not include some of the customizations I am asking for - but I do feel that allowing these customizations will improve the product for users who look to fit Kendo in their companies design systems.

File: scss/window/_layout.scss [hunk 10]
```diff
@@ -21,19 +28,22 @@
 
     // Title bar
     .k-window-titlebar {
-        padding: $window-titlebar-padding-y $window-titlebar-padding-x;
-        min-height: 16px;
```
IanMuirhead: @joneff I've removed the min-height here.

scss/window/_theme.scss
IanMuirhead: The only piece which doesn't fit with that is we are not currently setting a border-color on the window - I will add the variable  though?

scss/dialog/_theme.scss
```diff
@@ -1,19 +1,19 @@
 @include exports("dialog/theme") {
 
+    $dialog-titlebar-bg: $accent none !default;
+    $dialog-titlebar-text: contrast-wcag($dialog-titlebar-bg) !default;
```
IanMuirhead: @joneff I believe you requested this change.

scss/window/_theme.scss
```diff
@@ -4,6 +4,9 @@
     $window-shadow: 0 3px 3px 0 rgba(0, 0, 0, .06) !default;
     $window-focused-shadow: 1px 1px 7px 1px rgba(0, 0, 0, .3) !default;
 
+    $window-titlebar-border-color: inherit !default;
+    $window-titlebar-bg: $accent none !default;
+    $window-titlebar-text: contrast-wcag($window-titlebar-bg) !default;
```
IanMuirhead: @joneff I believe you requested this change.

scss/window/_layout.scss
joneff: What if we add titleTemplate or something similar in the window / dialog? That would provide parity with [Bootstrap modal component](http://v4-alpha.getbootstrap.com/components/modal/#modal-components) and will allow for greater customization.

IanMuirhead: Sounds good.  I'll update this morning!

IanMuirhead: @joneff How about something like this?

https://www.sassmeister.com/gist/2fc7f5c9bd25b6563ebaa70721273f62

scss/_variables.scss
```diff
@@ -99,8 +99,7 @@ $cell-padding: 8px !default;
 
 $window-inner-padding-x: 16px !default;
 $window-inner-padding-y: 16px !default;
-$window-titlebar-padding-x: 16px !default;
-$window-titlebar-padding-y: 16px !default;
+$window-titlebar-padding: 16px !default;
```
joneff: Let's go with the old -x / -y way for consistency

scss/window/_layout.scss
joneff: I actually meant to add titleTemplate property in the widget, so you can either have pure text or template

scss/window/_theme.scss
```diff
@@ -4,6 +4,9 @@
     $window-shadow: 0 3px 3px 0 rgba(0, 0, 0, .06) !default;
     $window-focused-shadow: 1px 1px 7px 1px rgba(0, 0, 0, .3) !default;
 
+    $window-titlebar-border-color: inherit !default;
+    $window-titlebar-bg: $accent none !default;
+    $window-titlebar-text: $accent-contrast !default;
```
joneff: Those should be tied to `$header-` variables, not `$accent`. You can customize them later

joneff (changes requested):

joneff:
@IanMuirhead, I think after you tweak the `-x / -y` vars and the window titlebar colors, we are all set.

/cc @gyoshev

scss/window/_layout.scss
gyoshev: The angular dialog [provides a custom titlebar](http://www.telerik.com/kendo-angular-ui/components/dialog/#toc-title). It can get the custom styling through it:

    <kendo-dialog>
        <kendo-dialog-titlebar>
             <span class="custom-title">{{ titleField }}</span>
        </kendo-dialog-titlebar>
    </kendo-dialog>

@IanMuirhead does this fit your design system?

IanMuirhead: @gyoshev My preference would always be to do it via the theme but that does look like we would have much more flexibility there without needing to push all of our requirements into the main theme.

I'll chat with my team and come up with a solution with that option.

IanMuirhead: @joneff Understood - I assume we're still okay with adding the `text-transform` and `font-family` to the theme?

scss/mixins/_decoration.scss
```diff
@@ -2,12 +2,12 @@
 @mixin fill( $color: null, $bg: null, $border: null, $gradient: null ) {
     border-color: $border;
     color: $color;
-    background-color: $bg;
+    background: $bg;
```
IanMuirhead: @gyoshev This wasn't in the original pull request, but I just realized that it would have introduced a bug if this didn't get changed as well. I'm not sure if this is something you would want to change / include in the theme - adds a lot of flexibility to the `@fill` mixin but this may introduce some issues where it is used elsewhere.

scss/dialog/_theme.scss
```diff
@@ -1,19 +1,19 @@
 @include exports("dialog/theme") {
 
+    $dialog-titlebar-border-color: inherit !default;
+    $dialog-titlebar-bg: $header-bg none !default;
```
IanMuirhead: @gyoshev This is one of the variables I pass to the `@fill` mixin.

scss/dialog/_theme.scss
```diff
@@ -1,19 +1,19 @@
...
 
     .k-dialog {}
 
 
     // Titlebar
     .k-dialog-titlebar {
-        color: $accent-contrast;
-        background-color: $accent;
-        background-image: none;
+        @include fill( $dialog-titlebar-text, $dialog-titlebar-bg, $dialog-titlebar-border-color );
```
IanMuirhead, line 12: @gyoshev Here is where we pass that variable in to the `@fill` mixin

IanMuirhead (commented):
@gyoshev Highlighted a couple of changes to the `@fill` mixin required.

scss/mixins/_decoration.scss
joneff: The short hands, like `background`, reset lots of stuff. That line, by itself means:

```scss
selector {
    background-color: $bg; // that's the only things that's set
    background-position: initial; // or 0 0 in older browsers
    background-repeat: initial; // or repeat in older browsers
    background-origin: initial; // or padding-box in older browsers
    background-image: initial; // or none in older browsers
    background-size: initial; // or auto in older browsers
}
```

That's why we specifically use the fully qualified properties.

joneff (changes requested):

IanMuirhead:
@joneff Swapped the background in the fill back to the original state.

joneff (approved):

joneff:
Seems good to me. Few minor variable naming nitpicks, but no big deal really.

@gyoshev ?

scss/window/_layout.scss
gyoshev: I still feel that some variables are too specific, as they are used just once and do not allow to be used anywhere else.

       font-size: $window-title-font-size;
       font-family: $window-title-font-family;
       text-transform: $window-title-text-transform;

If using the custom titlebar is not an option, have you considered overriding the styles in your design system, after importing the Kendo UI theme?

IanMuirhead: @gyoshev 

We for the most part try to not override or create custom CSS - it becomes quite fragile and we have been burned by the Kendo theme changing between our releases. I prefer the ability to get this concept into the theme and for it to be supported as there is less CSS for us to maintain within our system. 

Let me know what your preference is - I am happy to remove.  There would be still some wins for both sides if we remove this portion of it and we can figure out a solution on our end.

gyoshev: I think the variables are overly specific and should not be exposed in that format. I understand your requirements for the headers to be consistent, and perhaps we can introduce another way to achieve this, through another issue / PR. An idea that comes to mind is to expose a `@mixin` that can be redefined in consuming code, and is used across all component headers.

IanMuirhead: @gyoshev A mixin like the one I mocked up?

https://www.sassmeister.com/gist/2fc7f5c9bd25b6563ebaa70721273f62

IanMuirhead:
@gyoshev @joneff 

This should be good to go.

gyoshev:
Cheers! The variables will be available with the next release of the theme. You can [use the development channel](http://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) to get a build before it is released publicly, and test it out in your application.

Issue #575, opened 2017-06-30
MultiSelect clear button is not properly positioned


yordanov:
Clear button should be top-aligned, not vertically centered:

![multiselect-clear](https://user-images.githubusercontent.com/1118895/27731584-4fac275a-5d96-11e7-937e-119a7f88153e.png)

siyka-marinova:
Fixed in `@progress/kendo-theme-default@2.39.0`

PR #576, opened 2017-07-03
fix(multiselect): clear button position

inikolova:
Closes telerik/kendo-theme-default#575

### DIFF
--- a/scss/multiselect/_layout.scss
+++ b/scss/multiselect/_layout.scss
@@ -9,6 +9,11 @@
             visibility: hidden;
         }
 
+        .k-clear-value {
+            top: $padding-x-lg / 2;
+            margin-top: 0;
+        }
+
         &[dir='rtl'],
         .k-rtl & {
             li.k-button {


joneff (approved):

joneff:
@inikolova I've been thinking on this... the best way to position the clear button, in this particular case is to align it centrally in a box that's as big as the tokens and place the box the same distance from the ledge:

image

inikolova:
cc @yordanov

yordanov:
Yep, as soon as these dimensions are kept:

dimensions

joneff:
@yordanov we can't position the edge of an icon. Instead, we can position the icon container.

yordanov:
Icon's placeholder width x height is 16px  x 16px. As seen on the image above: 
Position of icon's **center** from **top**:  6px + 16/2px = **14px**
Position of icon's **center** from **right**: 6px + 16/2px = **14px**

Position of icon's placeholder top-right edge: top: 6px; right: 6px;

scss/multiselect/_layout.scss
```diff
@@ -9,6 +9,11 @@
             visibility: hidden;
         }
 
+        .k-clear-value {
+            top: $padding-x-lg / 2;
```
gyoshev, line 13: The value fixes the issue, but consider using a `y`-unit when calculating vertical offset. If left like this, changing the `padding-x-lg` will offset the icon vertically, which is not intuitive for the developers who use the theme.

gyoshev:
I agree with @joneff about changing the icon layout, as it may result in less styles overall and also allows to increase the button hit area (which is currently 16px x 16px, requires some precision). Since it is likely to require a rendering change it may take some time to propagate across the components, we can fix the issue with this PR and improve upon it in another.

@yordanov do you think that the hit area is adequate for the button?

yordanov:
Yep, the bigger the hit area - the better.  Token sized clear button looks fine to me.

Issue #476, opened 2017-05-03
Spreadsheet scroll-sheets buttons are not visible in IE


Dimitar-Goshev:
![image](https://cloud.githubusercontent.com/assets/6614183/25665930/c0b9aa68-3027-11e7-822f-f2073ba7b5cb.png)

joneff:
Fixed by 3fffc65

PR #582, opened 2017-07-04
fix(tabstrip): scroll buttons are not aligned in IE

joneff:
Fixes https://github.com/telerik/kendo-theme-default/issues/476


Dimitar-Goshev (approved):
buttons are now visible

PR #594, opened 2017-07-06
fix/tabstrip variable usage

IanMuirhead:
Assigns the selected variable to tabstrip items which are active

@joneff I think there was a bug in the variables which you used for the colors

### DIFF
--- a/scss/tabstrip/_theme.scss
+++ b/scss/tabstrip/_theme.scss
@@ -33,7 +33,7 @@ $tabstrip-content-border: $panel-bg !default;
             &.k-state-hover {
                 border-color: $tabstrip-item-hovered-border;
                 color: $tabstrip-item-hovered-text;
-                background-color: $tabstrip-item-hovered-border;
+                background-color: $tabstrip-item-hovered-bg;
             }
 
             &.k-state-active {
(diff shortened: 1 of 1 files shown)


scss/tabstrip/_theme.scss
```diff
@@ -62,30 +62,30 @@ $tabstrip-content-border: $panel-bg !default;
     .k-tabstrip-top {
         > .k-tabstrip-items {
             .k-item.k-state-active {
-                border-bottom-color: $tabstrip-item-selected-bg;
+                border-bottom-color: $tabstrip-item-selected-border;
```
joneff, line 65: that needs to be `-bg`, as well as the rest of the instances, so we can get the "enclosure" like appearance.

joneff (changes requested):

scss/tabstrip/_theme.scss
IanMuirhead: @joneff Totally missed this response - if it is required to be the 'bg' value here what's the point of having the selected border color?

IanMuirhead:
Going to close this PR.

I've created a solution locally which resolved our use-cases.

PR #613, opened 2017-07-12
Notification improvements

joneff:
Related: https://github.com/telerik/kendo-theme-bootstrap/issues/193
Fixes: https://github.com/telerik/kendo-theme-bootstrap/issues/149
Related: https://github.com/telerik/kendo-ui-core/issues/3323

### DIFF
--- a/scss/mixins/vendor/bootstrap/_functions.scss
+++ b/scss/mixins/vendor/bootstrap/_functions.scss
@@ -1,0 +1,13 @@
+// We need this before we update to the latest version
+// Modified from https://github.com/twbs/bootstrap/blob/v4-dev/scss/_functions.scss
+$theme-color-interval: 8% !default;
+@function theme-color-level($color, $level: 0) {
+    $color-base: if($level > 0, #000, #fff);
+
+    @if $level < 0 {
+        // Lighter values need a quick double negative for the Sass math to work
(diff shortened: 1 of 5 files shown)


scss/mixins/vendor/bootstrap/_functions.scss
```diff
@@ -0,0 +1,13 @@
+// We need this before we update to the latest version
+// Modified from https://github.com/twbs/bootstrap/blob/v4-dev/scss/_functions.scss
+$theme-color-interval: 8% !default;
+@function theme-color-level($color, $level: 0) {
```
gyoshev, line 4: It seems that this function is not used anywhere in our code. Why do we need it?

gyoshev (commented):
Renaming `$alert-*` to `$notification-*` seems appropriate, given that the colors are used only there. We could include this in the release notes to notify customers who might have depended on the internal variables. Not sure why the `theme-color-level` is needed, might need more explanation in the comments if it's hidden.

scss/mixins/vendor/bootstrap/_functions.scss
joneff: It's used in https://github.com/telerik/kendo-theme-bootstrap/commit/0683a67bcaf3fd02c88ad91a6e383eac48e5d1f4

I will create a separate PR for that

PR #616, opened 2017-07-12
fix(grid): prevent wraping and vertically align header column menu in IE

magdalenaan:
related https://github.com/telerik/kendo-ui-core/issues/3396

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -470,6 +470,11 @@
             margin: 0;
             margin-top: calc(#{$grid-form-component-offset} * -1);
             margin-bottom: calc(#{$grid-form-component-offset} * -1);
+            margin-right: (-$cell-padding-x);
+
+            & + .k-link {
+                margin-right: calc(#{$button-calc-size} - #{$cell-padding-x}); // .k-header-column-menu width
+            }
         }
 
         .k-header > .k-link > .k-icon {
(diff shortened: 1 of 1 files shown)


scss/grid/_layout.scss
```diff
@@ -468,8 +468,13 @@
         }
         .k-header-column-menu {
             margin: 0;
-            margin-top: calc(#{$grid-form-component-offset} * -1);
-            margin-bottom: calc(#{$grid-form-component-offset} * -1);
+            margin-top: (-$button-padding-y);
+            margin-bottom: (-$button-padding-y);
```
joneff: `$grid-form-component-offset` does perfect job of aligning components. Could you revise?

joneff (changes requested):

scss/grid/_layout.scss
magdalenaan: Actually nested calc() functions with nested "calc()" string does not work in IE, so this solution does not work in that browser. I set the "$button-padding-y" margin because of margin of  the ".k-link" next to the column menu has negative margin with the same value as its padding:

-$cell-padding-y + ( $cell-padding-y - $button-padding-y ) = -$button-padding-y
where the first one is margin-top of ".k-link", and the expression in the brackets is difference of ".k-link" and ".k-header-column-menu" top padding.

joneff (approved):

joneff:
We've optimized the formula for `$grid-form-component-offset`.

PR #619, opened 2017-07-14
fix(slider): extract variables for metrics

inikolova:


scss/slider/_layout.scss
```diff
@@ -259,16 +262,29 @@ $slider-track-thickness: 4px !default;
...
 
+        $slider-draghandle-offset: -(($slider-draghandle-size - ($slider-track-thickness / 2)) / 2) !default;
+        $slider-draghandle-active-offset: -(($slider-draghandle-active-size - ($slider-track-thickness / 2)) / 2) !default;
+
         .k-slider-horizontal & {
-            height: 14px;
-            top: -6px;
-            width: 14px;
+            height: $slider-draghandle-size;
+            top: $slider-draghandle-offset;
```
joneff: Could we go with `top: 50%; transform: translateY(-50%)` ?

inikolova: Yes, we could. Would you like to make this change?

scss/slider/_layout.scss
```diff
@@ -259,16 +262,29 @@ $slider-track-thickness: 4px !default;
...
+            }
         }
 
         .k-slider-vertical & {
-            height: 14px;
-            left: -6px;
-            width: 14px;
+            height: $slider-draghandle-size;
+            left: $slider-draghandle-offset;
+            width: $slider-draghandle-size;
```
gyoshev: seems we have the same width / height declarations for both vertical and horizontal handles, these can be moved to the draghandle in general.

scss/slider/_layout.scss
joneff: Sure

Issue #618, opened 2017-07-14
Missing styles for large ticks (k-tick-large)


inikolova:
Can be observed in the following demo (EQUALIZER): http://demos.telerik.com/kendo-ui/slider/index

PR #630, opened 2017-07-21
fix(slider): large ticks appearance

inikolova:
See telerik/kendo-theme-default#618


gyoshev:
Based on the number of occurrences of both classes in telerik/kendo, I think we can merge this :smile: 

image

theOrlin (approved):

PR #640, opened 2017-07-25
feat(timepicker): add styling for timepicker

inikolova:


scss/datetime/_layout.scss
```diff
@@ -1,4 +1,25 @@
+// Infinite calendar
+@mixin hide-scrollbar($dir: 'right') {
```
gyoshev: We have the same code in `calendar/_layout.scss`. Extract it to `mixins/core/` and reuse it, to reduce duplication. If the different `$max-scrollbar` value is needed, use a mixin parameter.

scss/datetime/_layout.scss
```diff
@@ -118,4 +139,95 @@
 
     }
 
+    // Infinite timepicker
+    // Header
+    .k-time-header { // Same as the calendar header
```
gyoshev: If the header is the same by design, perhaps we can use the `k-calendar-header` class? There are also duplicated styles in the _theme file, with k-time-now acting as k-today.

gyoshev (commented):
It seems that some of the calendar styling can be reused, which will trim down the styles and reduce duplication

scss/datetime/_layout.scss
inikolova: After further discussion with @ggkrustev , we decided to use `<button>` for the Now action in the timepicker, which differs from the Calendar component (where `<span>` is used). I will modify the styling in the timepicker header

gyoshev: Does the same reasoning apply to the calendar component? Using a `<button>` might be better for the calendar accessibility, too.

PR #655, opened 2017-07-31
Update buttons styling

inikolova:


scss/button/_layout.scss
```diff
@@ -232,6 +233,7 @@
         &::before {
             content: "";
             background-color: currentColor;
+            @include border-radius();
```
gyoshev, line 236: It seems the design of the default theme does not have a border-radius on the bare buttons ([link](https://github.com/telerik/kendo-ux/blob/master/themes/default/buttons.psd)). This may be intentional for some uses in Kendo UI for jQuery (MediaPlayer / ToolBar?). If so (@joneff can confirm), it's better to leave it as a variable.

joneff: Thus far, only theme default seems like an exception. We could override `border-radius` only in the theme file.

If at a later stage we deem necessary we can introduce a variable.

inikolova: Done!

PR #656, opened 2017-07-31
fix(timepicker): add time list focus state; jumping list items when sc…

inikolova:
…rolling

### DIFF
--- a/scss/datetime/_layout.scss
+++ b/scss/datetime/_layout.scss
@@ -135,6 +135,7 @@
     }
 
     $time-list-wrapper-padding: $padding-y * 8;
+    $time-list-wrapper-focused-border-width: 2px;
 
     // Content
     .k-time-list-wrapper {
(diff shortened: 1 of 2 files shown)


scss/datetime/_layout.scss
```diff
@@ -135,6 +135,7 @@
     }
 
     $time-list-wrapper-padding: $padding-y * 8;
+    $time-list-wrapper-focused-border-width: 2px;
```
gyoshev: No need for `-wrapper` in the variable names. I'd even go further and shorten the latter to `$time-list-focus-size`, which will also hide the implementation detail that it is achieved with `border` (we can change to `box-shadow` and the name will still make sense).

scss/datetime/_theme.scss
```diff
@@ -1,6 +1,7 @@
 @include exports("datetime/theme") {
     $date-focus-shadow: $list-item-focused-shadow !default;
     $time-highlight-shadow: inset 0 0 0 1px rgba(0, 0, 0, .13) !default;
+    $list-wrapper-focused-border-color: rgba(0, 0, 0, .13) !default;
```
gyoshev: Name suggestion: `$time-list-focus-color`

PR #662, opened 2017-08-03
fix(timepicker): update timelist focus state

inikolova:


scss/datetime/_theme.scss
```diff
@@ -1,7 +1,7 @@
 @include exports("datetime/theme") {
-    $date-focus-shadow: $list-item-focused-shadow !default;
-    $time-highlight-shadow: inset 0 0 0 1px rgba(0, 0, 0, .13) !default;
-    $time-list-focus-color: rgba(0, 0, 0, .13) !default;
+    $time-highlight-color: rgba(0, 0, 0, .13) !default;
+    $timelist-focused-bg: rgba(0, 0, 0, .04);
+    $focused-title-color: #000 !default;
```
gyoshev: I believe `$time-list-highlight-color`, `$time-list-focused-bg` and `$time-list-title-focus` will be better names, matching the prefix we have in `_layout`.

Issue #602, opened 2017-07-10
The NumericTextBox arrows are misplaced when the widget is used in the Grid(Only in Edge)


simonssspirit:
The NumericTextBox arrows are misplaced when the widget is used in the Grid(Only in Edge).

The issue can be observed for the Default v2 and the Bootstrap v4 themes:

'http://demos.telerik.com/kendo-ui/grid/editing'

PR #664, opened 2017-08-08
fix(numerictextbox): Adjust flex of input fixing Edge weirdness

theOrlin:
Fixes problems with NumericTextBox spinners in Edge browser. Dealing with the following issues:

- https://github.com/telerik/kendo-theme-bootstrap/issues/295
- https://github.com/telerik/kendo-theme-default/issues/602
- 9 in https://github.com/telerik/kendo-theme-bootstrap/issues/218


joneff (approved):
It seems like we should change all instances from `flex: 1` to `flex: 1 1 0`...

PR #666, opened 2017-08-10
Ripple effect, floating label, checkbox / radiobutton / math improvements

gyoshev:
Some features that are necessary for the Material theme. It is all nested in one PR due to the intertwined features (ripple effects, input sizing). Goes hand-in-hand with the [`checkbox-styling` branch in theme-material](https://github.com/telerik/kendo-theme-material/tree/checkbox-styling).

* `add-two` and `add-three` are fixed for zero units (`calc( 0 + 2px )` is invalid)
* adds ripple effect styles for all themes
* checkboxes and radio buttons now use `::before` for their box, and `::after` for the indeterminate/checked states. This allows the checked state to be animated, and for the styling to be consistent. There are also variables that allow [configuration of the checkbox size, and the style of the indeterminate state](https://github.com/telerik/kendo-theme-default/blob/material-checkbox/scss/input/_layout.scss#L146-L149) (dash or square).
* these commits [imports the floating label styles from material](https://github.com/telerik/kendo-theme-material/pull/3) and makes it available in all themes.

@mvgmvg I had to [revert the checkbox hover commit](https://github.com/telerik/kendo-theme-default/commit/bae6fb993339f5b7177b0f3c06337788e3871044) and [rephrase it](https://github.com/telerik/kendo-theme-default/commit/970f0bf108d7065aed42f9ac49fa3200529b6ed2), please verify if the bugfix is still in place.

/cc @theOrlin @inikolova

### DIFF
--- a/scss/all.scss
+++ b/scss/all.scss
@@ -3,6 +3,7 @@
 
 // Dependencies
 @import "common/all";
+@import "ripple";
 @import "popup";
 
 
--- a/scss/input/_layout.scss
+++ b/scss/input/_layout.scss
@@ -114,8 +114,8 @@
             left: $checkbox-size / 2;
             right: auto;
             bottom: auto;
-            width: $checkbox-size * 2;
-            height: $checkbox-size * 2;
+            width: $checkbox-size * 5 / 2;
+            height: $checkbox-size * 5 / 2;
             transform: translate(-50%, -50%);
             border-radius: 50%;
         }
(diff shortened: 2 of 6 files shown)


magdalenaan (commented):
I have tested the branch regarding to mentioned conflicts with a previous checkbox states fixing commit - all states are correct.

joneff (approved):

PR #670, opened 2017-08-11
feat(forms): integration with Bootstrap form-control-sm / lg

inikolova:
…sm and form-control-lg classes

Related to https://github.com/telerik/kendo-angular/issues/698

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -290,7 +290,6 @@
 
     $bootstrap-integration-input-height-lg: ($twbs-input-padding-y-lg * 2) + ($twbs-font-size-base-lg * $twbs-input-line-height-lg) !default;
 
-    $double-input-border-width: $input-border-width * 2;
     $numerictextbox-icon-offset-sm: 2px !default;
 
     .k-widget.form-control {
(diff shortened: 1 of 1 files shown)


scss/common/_forms.scss
```diff
@@ -274,6 +274,25 @@
...
+    $bootstrap-integration-input-height-sm: ($twbs-input-padding-y-sm * 2) + ($twbs-font-size-base-sm * $twbs-input-line-height-sm) !default;
+
+    $twbs-font-size-base-lg: 1.25rem !default;
+    $twbs-input-padding-x-lg: 1rem !default;
+    $twbs-input-padding-y-lg: .5rem !default;
+    $twbs-input-line-height-lg: 1.5 !default;
+
+    $bootstrap-integration-input-height-lg: ($twbs-input-padding-y-lg * 2) + ($twbs-font-size-base-lg * $twbs-input-line-height-lg) !default;
+
+    $double-input-border-width: $input-border-width * 2;
```
gyoshev: This variable can be inline in the `add-two` call, so that

     add-two($bootstrap-integration-input-height, $double-input-border-width);

becomes

     add-two($bootstrap-integration-input-height, 2 * $input-border-width);

There is more repetition, but it is clearer than introducing a new variable which requires lookup.

gyoshev (approved):
Looks good, just a small nitpick. Might be a good idea to add a section to the [Bootstrap integration help topic](http://www.telerik.com/kendo-angular-ui/components/forms/integration-with-bootstrap/) that shows these in action.

joneff (approved):

PR #682, opened 2017-08-18
fix(button): outline primary buttons appearance

inikolova:


scss/button/_theme.scss
```diff
@@ -1,4 +1,6 @@
 @include exports( "button/theme" ) {
+    $outline-primary-button-text: $accent !default;
+    $outline-primary-button-border: $accent !default;
```
gyoshev: Shouldn't these always be the same, by design? If so, it may be more convenient to set only the text color and use `currentColor` for the borders. This way customers can change the button behavior by just setting the button `color`.

joneff (approved):

Issue #600, opened 2017-07-10
Grid, grouping does not work


DominikJaniec:
# Bug: Groups reorder does not work

Using this Dojo: http://dojo.telerik.com/OXaVa/6
You could see how grouping does not work (or sometimes works) like on attached GIF.
![not_so_groupable](https://user-images.githubusercontent.com/3832797/28022053-3be78fd8-658a-11e7-8a5d-c8ce50a9c45a.gif)
GIF is from Chrome, but this effect is also in Firefox or IE.

In fact problem is with impossibility (or lack of luck) to select certain positions at Groups' header.
I'm using SASS version of yours theme, but at Dojo is same behavior like in my case.

----
Also I would like to point that there is no way to move column's ordinary header to the last position. You have to drop it at next-to-last position and then drag the last one to the left. Additionally "drop indicators" for columns reorder are at bottom of the page, instead of between target Headers.
But I think you are aware of it: https://github.com/telerik/kendo-ui-core/issues/3250

gyoshev:
Thank you for reporting this. It appears the issue is caused by the theme, as the problem does not occur with other themes. Requires further investigation on why it happens.

joneff:
@DominikJaniec @gyoshev the issue is caused by missing styles for grid column reordering. When hovering adjacent columns, the reorder indicators appear below the grid and take ever so slightly more space (about 1em or 1rem), hence the scrolling.

That should be a straightforward fix.

joneff:
Needs https://github.com/telerik/kendo-theme-default/pull/696

matanasov:
@joneff 
The main issue is still not resolved - drag a column to the grouping toolbar and drop it **between other columns**. Note that drag-drop **in the first/last** position seems to work as expected

matanasov:
@theOrlin 
The issue is still valid:
1. Group by `Order Date`
2. Add `Ship Country` grouping **after** `Order Date`
3. Add `Ship City` grouping **after** `Ship Country`
4. Add `Ship Name` grouping **as a middle group** - it is not added!

PR #696, opened 2017-08-29
Add styles for reorder cue

joneff:
Related to https://github.com/telerik/kendo-theme-default/issues/600


scss/common/_dragclue.scss
```diff
@@ -1,23 +0,0 @@
-@include exports( "common/dragclue" ) {
-
-    // Layout
-    .k-drag-clue {
```
gyoshev, line 4: These styles seem to be still used in the TreeList, TreeView, and Grid:

image

scss/grid.scss
```diff
@@ -10,4 +10,4 @@
 
 // Component
 @import "grid/layout";
-@import "grid/theme";
+@import "grid/theme";
```
gyoshev, line 13: The opposite is specified by the [.editorconfig file](https://github.com/telerik/kendo-theme-default/blob/develop/.editorconfig#L8). Make sure that any used editor [supports it or has a plugin installed](http://editorconfig.org/#download). The reasoning for having these is described in [this SO thread](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

gyoshev (changes requested):

scss/common/_dragclue.scss
joneff: the styles are not removed. just singled out. They already exist in base: https://github.com/telerik/kendo-theme-default/blob/647f69552b6ebe8d4aecb6a484e8615ce19f11bc/scss/common/_base.scss#L596-L618

joneff (commented):
The styles already exist in common/base

scss/grid.scss
joneff: As soon as I merge the other request, I will fix this one as well

gyoshev (approved):

PR #698, opened 2017-08-29
Enforce editorconfig

gyoshev:
These changes start enforcing the rules [described in the .editorconfig file](https://github.com/telerik/kendo-theme-default/blob/develop/.editorconfig) and fix any violations.

Make sure that any editor you use [either supports editorconfigs or has a plugin installed](http://editorconfig.org/#download).

/cc @inikolova @mvgmvg @theOrlin


joneff (approved):
Based on the SO thread on the matter, we'll go with EOL@EOF

Issue #704, opened 2017-08-31
No data template not positioned properly in IE


dtopalov:
The DropDowns components No Data template is not vertically centered in IE like it is in other browsers. Reproducible in the online demos:

[http://www.telerik.com/kendo-angular-ui/components/dropdowns/combobox/templates/#toc-no-data-template](http://www.telerik.com/kendo-angular-ui/components/dropdowns/combobox/templates/#toc-no-data-template)

Related to the following IE bug:

[https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview#tabs](https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview#tabs)

Workarounds:

    .k-popup { 
  	  display: flex;
  	}
or

    .k-nodata{ 
  	  height: 138px;
  	}

siyka-marinova:
Fixed in `@progress/kendo-theme-default@2.41.1-dev.201709051522`

siyka-marinova:
Released in `@progress/kendo-theme-default@2.41.1`

PR #709, opened 2017-09-01
fix(popup): no data template alignment in IE

inikolova:
See https://github.com/telerik/kendo-theme-default/issues/704


scss/popup/_layout.scss
```diff
@@ -1,6 +1,7 @@
 @include exports('popup/layout') {
 
     .k-popup {
+        display: flex;
```
joneff, line 4: we should ensure `flex-direction: column` and possibly `align-items: stretch`.

joneff (changes requested):

joneff (approved):

PR #716, opened 2017-09-05
fix(grid): Restore missing filter row Autocomplete border

theOrlin:
Fixes missing left border of Autocomplete widget in filter row.

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -551,6 +551,10 @@
             .k-numerictextbox {
                 width: auto;
             }
+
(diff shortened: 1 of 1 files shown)


joneff (changes requested):
You are looking in the wrong direction. Reason for this is the class `.k-header`, which contributes to the border being gone and the rest of the borders having a slightly different color.

The solution is a work in progress in https://github.com/telerik/kendo/tree/remove-k-header-from-pickers and as soon as we finish there, it will be updated here.

PR #726, opened 2017-09-13
fix(animation): remove zoom & push translation to support single element animation

kirchoni:
When animating multiple child-containers the translation was needed to stack them together. This
prevented a single element to be animated in/out because the translation made it disapear outside
the parent-container instantly. In addition, some tweaks were added to zoom animations.

### DIFF
--- a/scss/common/_animations.scss
+++ b/scss/common/_animations.scss
@@ -28,17 +28,12 @@
         }
 
         &-exit {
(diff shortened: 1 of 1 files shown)


mitagg:
Giving the fact that this functionality is only affecting the react packages I do approve this pull request. 

cc. @joneff @gyoshev

PR #733, opened 2017-09-22
docs: add clarification

dmihaylo:


dtopalov (changes requested):
Default and Bootstrap themes are supposed to be supporting all components. Clarification is necessary for the Material theme only.

PR #749, opened 2017-10-06
fix(grid): Add scrollbar to filter menu to avoid IE/Edge stretch bug …

theOrlin:
…(telerik/kendo-ui-core#2898)

Dealing with https://github.com/telerik/kendo-ui-core/issues/2898


theOrlin:
As per @hadjieva, this is not an acceptable solution to this. Closing until a different solution is found.

PR #757, opened 2017-10-18
Grid filter menu

gyoshev:
| Before | After (no rendering changes) | After (with added classes) |
|-|-|-|
| image | image | image |

To enable the rightmost styling:

1. add the `k-filter-menu-container` class to the nameless wrapper div inside the column menu
2. add the `k-action-buttons k-button-group` class to the nameless div that holds the buttons
3. flip the buttons to make the rendering the same as in the timepicker / spreadsheet

image

This PR also contains fixes for the datepicker and multiselect, visible when nested in the filter menu.

Do not squash.

### DIFF
--- a/scss/action-buttons/_theme.scss
+++ b/scss/action-buttons/_theme.scss
@@ -1,6 +1,7 @@
 @include exports('action-buttons/theme') {
 
     .k-action-buttons {
+        border-color: $popup-border;
 
         .k-button {
             color: inherit;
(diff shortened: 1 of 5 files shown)


joneff (approved):

joneff:
@theOrlin @ag-petrov it seems that the change needs to be applied on our end as well

gyoshev:
Btw, I've added a `BACKCOMPAT` tag to styles and rules that can be removed after this hits production. We can keep them for a while until customers who use the packages directly upgrade (one or two releases time).

PR #758, opened 2017-10-18
fix(switch): improve switch labels position telerik/kendo-ui-core#3229

magdalenaan:
related https://github.com/telerik/kendo-ui-core/issues/3229

### DIFF
--- a/scss/switch/_layout.scss
+++ b/scss/switch/_layout.scss
@@ -76,7 +76,7 @@ $switch-handle-size: (.8 * $switch-size) !default;
     }
 
     .k-switch-label-on {
-        left: calc(-1*(#{$switch-size}) + (0.2*#{$switch-size}));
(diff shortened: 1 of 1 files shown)


gyoshev (commented):
The switch now looks great! :+1: 

`0.8` seems to be an oddly specific multiplier, and I suspect it is related to the default `$switch-size`. How is it calculated?

magdalenaan:
It comes from shifting left with $switch-size and shifting right with spacing around the switch button:
`-1*$switch-size + ($switch-size - $switch-handle-size) = -1*$switch-size + 0.2*$switch-size = 0.8*$switch-size`
where `$switch-handle-size: (.8 * $switch-size)`
Actually in this case the `0.8 * $switch-size` could be replaced by `$switch-handle-size`.

joneff (approved):

PR #759, opened 2017-10-18
chore: parallelize build

gyoshev:
makes the local build twice as fast.

before: npm test  42.81s user 2.55s system 104% cpu **43.516 total**
after:  npm test  41.84s user 2.66s system 209% cpu **21.196 total**

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "build": "webpack --optimize-minimize --bail",
     "watch": "webpack --watch",
     "embed-assets": "node build/embed-assets.js",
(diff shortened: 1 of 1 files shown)


joneff (approved):

gyoshev:
It appears that this approach is [frowned upon by Travis](https://docs.travis-ci.com/user/common-build-problems/#My-build-script-is-killed-without-any-error), and it [kills the builds](https://travis-ci.org/telerik/kendo-theme-default/builds/291976054#L1178). This alone does not suffice as a reason to move the builds to our infrastructure, so let's put it on hold for now.

Issue #738, opened 2017-09-27
Loading indicator missing in latest Chrome


Zaixu:
With the latest version of Chrome 61.0.3163.100 (64-bit). The loading icon is no longer visible.

Just open [http://www.telerik.com/kendo-angular-ui/components/styling/icons/#](http://www.telerik.com/kendo-angular-ui/components/styling/icons/) (Unfortunately the plunker function does not work, but just open the example and you can see that any loading icons which is 16px no longer shows. 
17px will however, so seems like a size mismatch calculation.

gyoshev:
@Zaixu Thank you for reporting this. It appears that the issue is caused by a rounding error, given that the border-width is set through `em` units.

Until this is fixed, you can workaround it with the following styles:

    .k-i-loading:after, .k-i-loading:before {
        border-width: 0.05em;
    }

gyoshev:
Appears to be fixed in `v2.43.2`

PR #768, opened 2017-10-24
fix(loading): small icon is invisible in Chrome 61

gyoshev:
see telerik/kendo-theme-default#738


joneff (approved):

joneff:
The problem remains if font size is small enough that we don't get 1px border.

gyoshev:
Yes, this change sets the threshold to `font-size: 10px`. I think it is safe to say that we don't support that by default. And the workaround (to set `border-width`) is trivial in that case.

PR #770, opened 2017-10-24
fix: make drag handle to be visible when cells are frozen

magdalenaan:
related https://github.com/telerik/kendo-ui-core/issues/3254


joneff (approved):

gyoshev:
@mvgmvg set the scope of the fix to generate better release notes:

```
fix(spreadsheet): make drag handle always visible

... more details
```

PR #772, opened 2017-10-26
fix(icons): use horizontal flipping for specifc rtl icons

joneff:


gyoshev (approved):
The styles have no effect in Angular, so they are OK on our side. It is likely that we'll need to nest these in a `[dir=rtl]` selector to apply them for Angular components.

theOrlin (approved):

joneff:
@gyoshev good point. I will probably prepare another pull request

Issue #779, opened 2017-11-06
Grid does not resize its content properly if virtual scrolling is enabled.


MilenaCh:
1.Navigate to:
https://dojo.telerik.com/OYisU/11
2.Increase the width of some column

Result:
https://www.screencast.com/t/jbTFTMoaTH

PR #780, opened 2017-11-07
fix(grid): Fix column resizing with virtual scrolling enabled

theOrlin:
Dealing with https://github.com/telerik/kendo-theme-default/issues/779

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -754,6 +754,10 @@
         position: relative;
         border-style: solid;
         border-width: 0 1px 0 0;
+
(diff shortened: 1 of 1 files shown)


gyoshev:
It seems to me that adding these styles [breaks the locked column functionality](https://dojo.telerik.com/@gyoshev/OveXOf), as the two containers are displayed one below the other:

image

The same problem [does not appear](http://plnkr.co/edit/aTVnuCPrm2BMibFey8sH?p=preview) in Angular. Perhaps there is another solution?

gyoshev (approved):
Looks good with the Angular demos.

PR #783, opened 2017-11-09
Flexbox splitter

gyoshev:
Styles for the Angular implementation Splitter, and a fix for the splitter cursors in general.

### DIFF
--- a/scss/splitter/_layout.scss
+++ b/scss/splitter/_layout.scss
@@ -1,4 +1,5 @@
 @include exports("splitter/layout") {
+    $splitter-splitbar-size: 8px !default;
     $splitter-resize-icon-size: 10px !default;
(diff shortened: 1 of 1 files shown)


scss/splitter/_layout.scss
```diff
@@ -131,4 +132,45 @@
...
+
+            &[dir=rtl] {
+                flex-direction: row-reverse;
+            }
+        }
+        &.k-splitter-vertical {
+            flex-direction: column;
+
+            &[dir=rtl] {
+                flex-direction: column-reverse;
```
joneff: That should be removed. The rest is fine

joneff (approved):

scss/splitter/_layout.scss
gyoshev: Very nice catch, thanks!

PR #789, opened 2017-11-16
chore: add integrity tests

joneff:
Fixes: telerik/kendo/issues/7847
Related: telerik/kendo-theme-default/issues/720

@gyoshev  Should we include `karma start` in test or lint npm tasks? Perhaps we need another one for test-watch


gyoshev:
We usually have two tasks, one inside the [`npm test` script](https://github.com/telerik/kendo-theme-default/blob/develop/package.json#L39) and one for development, e.g. `npm run watch-test` (like in Angular and React), with `singleRun: false`.

joneff:
Point taken. I will update shortly

package.json
```diff
@@ -36,9 +36,14 @@
...
-    "test": "npm run lint && npm run build && npm run api-check && npm run twbs-compat",
+    "test": "npm run lint && npm run build && npm run api-check && npm run twbs-compat && npm run ts-single && npm run karma-single",
     "twbs-compat": "webpack --env.twbs-compat --bail",
-    "semantic-release": "semantic-release pre && semantic-prerelease publish && semantic-release post"
+    "semantic-release": "semantic-release pre && semantic-prerelease publish && semantic-release post",
+    "watch-test": "npm run ts-watch & npm run karma-watch",
+    "ts-single": "./node_modules/.bin/tsc --project ./tests/",
+    "ts-watch": "./node_modules/.bin/tsc --watch --project ./tests/",
+    "karma-watch": "karma start --auto-watch --no-single-run",
+    "karma-single": "karma start --no-auto-watch --single-run --browsers PhantomJS"
```
gyoshev: If I understand correctly, tests will run in PhantomJs as part of the theme verification, and with Chrome and PhantomJS locally. It seems that if we use only [Chrome headless](https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai) this may be simplified, and will require one less runner.

package.json
```diff
@@ -68,16 +73,25 @@
...
+    "karma": "^1.7.1",
+    "karma-chrome-launcher": "^2.2.0",
+    "karma-phantomjs-launcher": "^1.0.4",
+    "karma-qunit": "^1.2.1",
     "mime": "^1.3.4",
+    "qunitjs": "^2.4.1",
     "sass-lint": "^1.7.0",
     "sassdoc": "^2.1.20",
     "semantic-release": "^6.3.6",
+    "typescript": "^2.6.1",
```
gyoshev: I don't see much justification for using TypeScript in the tests so far, apart from type completion for the `metrics` object. Perhaps we can shave off some of the dependencies and configuration boilerplate if we use JS?

tests/button-qunit.ts
```diff
@@ -0,0 +1,16 @@
+QUnit.module("Button size", {
+    before: function() {
+        TestHelper.loadFixture( "/base/tests/fixtures/button.html" );
+    },
+    after: function() {
+        //kendo.destroy(QUnit.fixture);
```
gyoshev: Can be safely deleted.

.config/karma.conf.js
```diff
@@ -0,0 +1,84 @@
...
+
+
+        // list of files / patterns to load in the browser
+        files: [
+            "node_modules/jquery/dist/jquery.js",
+            "tests/lib/helpers.js",
+            "dist/all.css",
+            "tests/fixtures/**/*.html",
+            "tests/data/metrics.js",
+            "tests/**/*-qunit.js"
```
gyoshev, line 20: I'm assuming that in the different themes, the tests and fixtures will be reused with a different metrics object, is that correct? If so, what will be the configuration for that?

tests/lib/helpers.ts
```diff
@@ -0,0 +1,101 @@
+/// <reference path="../data/metrics.ts" />
+
+interface Window {
+    jQuery: JQueryStatic
+    QUnit: QUnit
+}
+
+class TestHelper {
```
gyoshev: It appears that the [Karma html2js preprocessor](https://github.com/karma-runner/karma-html2js-preprocessor) allows fixtures to be loaded directly by Karma. Does it fit the use-case?

joneff:
@gyoshev i've removed the TS dep and cleaned up the code. Also, I switched to ChromeHeadless per suggestion.

On html2js, the simple fixture loading is working, thus far. Should it fail to meet any of our cases, we'll switch to and alternative.

On multiple theme setup: yes, we are going to have multiple config files and we are going to feed the file via a param. However, I'd like to see how we go with the theme repo restructure first.

Issue #752, opened 2017-10-12
Window titlebar should be lower


magdalenaan:
### Bug report
When font-size is 14 pixels, window title-bar should be 46 pixels height instead of 63 pixels.

### Reproduction of the problem
1. Go to http://demos.telerik.com/kendo-ui/window/index
2. Set the "Default v2" theme

### Current behavior
https://www.screencast.com/t/dc0wLVpS

### Expected/desired behavior
https://www.screencast.com/t/nnZrcRqVQ44l

PR #790, opened 2017-11-16
Fixes/window title height

magdalenaan:
related https://github.com/telerik/kendo-theme-default/issues/752

### DIFF
--- a/scss/window/_layout.scss
+++ b/scss/window/_layout.scss
@@ -37,7 +37,7 @@
     .k-window-title {
         margin: 0;
         font-size: $font-size-lg;
-        line-height: 20px / $font-size-lg;
+        line-height: 1;
         text-overflow: ellipsis;
         overflow: hidden;
         cursor: default;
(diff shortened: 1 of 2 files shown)


scss/window/_layout.scss
```diff
@@ -53,6 +53,11 @@
         flex-shrink: 0;
         align-items: center;
         vertical-align: top;
+
+        .k-button {
+            margin-top: calc(#{-1*$button-padding-y} - 1px);
+            margin-bottom: calc(#{-1*$button-padding-y} - 1px);
```
gyoshev: `add-two($button-padding-y, 1px, -1)` might be cleaner here. Is there a variable that has the same meaning as the `1px`, for example the button border width, or the titlebar border width?

Please drop me a note (or assign me for review) when the styles affect an Angular component.

joneff:
@gyoshev I've applied the suggested changes. I've moved them to the containing element, as it seems more appropriate.

Issue #807, opened 2017-12-04
[Dialog - IE11] - Bug when height of dialog exceeds window height


IanMuirhead:
**Browser**: IE 11

When a dialog has very long text, in Chrome and other browsers the buttons remain at the bottom and the content area becomes scrollable.

In IE11 the content area does not become scrollable and it pushes the buttons outside of the viewable area.

See below:
http://plnkr.co/edit/ity8qHCfDz9gIapKxsYJ?p=preview

IanMuirhead:
@gyoshev Any thoughts on this one?

When setting a `max-height` to `k-window-content` the height no longer exceeds the window.

IanMuirhead:
Scratch that - got a fix @gyoshev ... will submit a PR.

IanMuirhead:
Fixed with #814

PR #814, opened 2017-12-05
fixes/ie-11-dialog-window-height

IanMuirhead:
- adds height to the `k-window` wrapping class which resolves an issue with IE11 not properly setting the height of the dialog when the text exceeds the display

### DIFF
--- a/scss/dialog/_layout.scss
+++ b/scss/dialog/_layout.scss
@@ -4,6 +4,7 @@
         display: flex;
         align-items: center;
         justify-content: center;
+        flex-direction: column;
         position: fixed;
         top: 0;
         left: 0;


IanMuirhead:
REFS: #807 

@gyoshev

gyoshev:
I am afraid that this style will set a 100% height to all windows, which is not intended.

It appears that adding `flex-direction: column` to the `.k-dialog-wrapper` element fixes the problem and still allows the window to automatically resize. Please [check this plunker out](http://plnkr.co/edit/jOeoZyuFkdVnhYbAlBua?p=preview) if it works in your scenario.

IanMuirhead:
@gyoshev 

Good point - was only testing on dialogs with the height beyond the window.

Will update the PR with your fix.

gyoshev (approved):

joneff (approved):

theOrlin (approved):

PR #820, opened 2017-12-08
fix(splitter): fix rtl mode reverse panes style

elena-gancheva:
see telerik/kendo-angular#1139


gyoshev (approved):

gyoshev:
Better comment: `fix(splitter): reverse panes in RTL mode`

PR #823, opened 2017-12-14
fix(forms): add k-state-invalid selector for color representation for…

kirchoni:
… invalid state

### DIFF
--- a/scss/common/_forms.scss
+++ b/scss/common/_forms.scss
@@ -256,7 +256,7 @@
         right: 3px;
     }
 
-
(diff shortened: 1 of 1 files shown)


mitagg (approved):
LGTM 👍  @gyoshev has the final word

mitagg:
Once we release this change, probably a subsequent release of the other themes will be required.

kirchoni:
As far as i can tell, the entire _forms.scss file is being imported in the bootstrap theme as it is.

PR #840, opened 2018-01-02
fix(combobox): imrpove combobox button width

magdalenaan:
related https://github.com/telerik/kendo/issues/8005


joneff:
The square buttons are theme-specific.

Rejecting.

PR #853, opened 2018-01-15
fix(inputs): fix input not having default font-size set

elena-gancheva:
see telerik/kendo-angular#1178


scss/input/_layout.scss
```diff
@@ -13,6 +13,7 @@
     $checkbox-dash-offset-x: $checkbox-border-width * 2 !default;
     $checkbox-dash-offset-y: ( $checkbox-size - $checkbox-dash-height ) / 2 !default;
 
+    $textbox-default-font-size: $font-size !default;
```
gyoshev: I'd remove the `-default` from the name. Indeed, there's `$textbox-default-width`, but that name isn't perfect either.

PR #866, opened 2018-01-31
Various fixes

joneff:
Needed for the add-tests branch

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -101,6 +101,8 @@ $window-inner-padding-x: 16px !default;
 $window-inner-padding-y: 16px !default;
 $window-titlebar-padding-x: 16px !default;
 $window-titlebar-padding-y: 12px !default;
+$window-titlebar-font-size: $font-size-lg !default;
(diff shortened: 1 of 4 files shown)


scss/_variables.scss
```diff
@@ -101,6 +101,8 @@ $window-inner-padding-x: 16px !default;
 $window-inner-padding-y: 16px !default;
 $window-titlebar-padding-x: 16px !default;
 $window-titlebar-padding-y: 12px !default;
+$window-titlebar-font-size: $font-size-lg !default;
+$window-titlebar-line-height: $line-height / (unitless( $font-size-lg ) / unitless($font-size) ) !default;
```
joneff, line 105: @gyoshev that unitless part is needed due to twbs-compat task. Bootstrap uses rem units and we use pixels. In addition, `$font-size` is not overwritten, hence we get invalid `px/rem`... Perhaps, we could add a twbs-compat-map.scss, to handle the vars that we have and bootstrap doesn't...

gyoshev:
PR moved to monorepo.

PR #875, opened 2018-06-13
docs: update legal entity

dmihaylo:


dmihaylo:
update moved to the kendo-themes monorepo

